PR #17, opened 2015-02-06
Fixes wrong attribute-selector in TS example

phbergsmann:


fsuter:
Thanks for the correction!

fsuter:
Forward-ported to 7 in 04ba8ad0333295810562e9a9e57688d8120ed87f

PR #19, opened 2015-03-29
[BUGFIX] Removed wrong stdWrap from *RootPaths

frans-beech-it:
partialRootPaths and layoutRootPaths do not have stdWrap support


marble:
Trusting you. Do we need to apply that for latest as well?

frans-beech-it:
Is also for master/latest

PR #20, opened 2015-04-20
Host cannot be detected using REMOTE_ADDR

luberti:
I believe there is a mistake in the way the condition on remote host is described.


marble:
Thank you! I'll apply the change to 'latest' to.

PR #21, opened 2015-04-29
[TASK] Mention parsetime output for config.debug

mbrodala:
config.debug does not only affect menu content objects but also parsetime output.


Chrissitopher:
Thanks for the contribution, Mathias!

PR #24, opened 2015-06-23
fix example for Condition "dayofyear"

davidsteeb:
fixed a small error in Conditions-Reference for "dayofyear" example. If the first day of the year (January, 1st) is 0 than [dayofyear = 5] will be true on January 6th, not [dayofyear = 7]


marble:
Thanks for spotting the error! Keep contributing : - )

PR #35, opened 2015-08-07
[BUGFIX] Properly document GIFBUILDER edge effect

yabawock:
The „edge“ transform offered by ImageMagick and GraphicsMagick performs edge detection on the image, not rounded corners.

Resolves: #26065


Chrissitopher:
Hi Morton,

thanks again!

PR #36, opened 2015-08-07
[TASK] Add more documentation to GIFBUILDER solarize effect

yabawock:
Add a short explanation of what solarize does, document what the value for the effect means.

Resolves: #26064


Chrissitopher:
Hi Morton, 

thanks for your contribution!

PR #43, opened 2015-09-22
Update Index.rst

ryanklarhoelter:
"[GLOBAL]" is for special purposes and shouldn't be used in normal cases. Conditions should be closed with "[END]" instead. "[GLOBAL]" should be documented separately.

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -22,7 +22,7 @@ available conditions see below.
 "[ELSE]" is available as else operator. It is a condition, which will
 return TRUE, if the previous condition returned FALSE.
 
-Each condition block is ended with "[GLOBAL]".
+Each condition block is ended with "[END]".
 
 
 Example:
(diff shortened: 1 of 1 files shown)


marble:
Hello Ryan, can you describe that special purpose? What the difference is? AFAIR that "[global]" has been used in many examples. If now we recommend "[END]" from now on I wouldn't do so without explaining the difference between "[global]" and "[end]".

ryanklarhoelter:
"[GLOBAL]" is a condition by itself that always returns "true". It ensures that the following TypoScript code is located in the global scope. So you can be sure that it's not affected by previous TypoScript (for example if anyone has forgotten to close a bracket). The Template Analyzer shows this very well: TYPO3 places a "[GLOBAL]" condition at the beginning of each TypoScript file. As a developer you can use "[GLOBAL]" for testing purposes (to ensure that your own condition works like expected). It's partly documented under https://docs.typo3.org/typo3cms/TyposcriptSyntaxReference/Syntax/Conditions/Index.html#the-global-condition too and further in Patrick Lobachers integrator book (german only; don't know if it's allowed to cite it here).

marble:
Great, thank you. I will adapt your explanation and fill it in (soon).

marble:
Now, here you go:
https://docs.typo3.org/typo3cms/TyposcriptReference/0.0/Conditions/Index.html#general-syntax
Good work!

ryanklarhoelter:
Great, thanks!

PR #49, opened 2015-11-17
[BUGFIX] Fix missing char in variables section

jousch:
Make TYPO3 integrators happy again


marble:
Thanks! Gimme more :-)

PR #50, opened 2015-11-20
Declare align deprecated with IMGTEXT and CTABLE

elmar-hinz:
It's not needed any more. Style should not be part of HTML.

### DIFF
--- a/Documentation/DataTypes/Align/Index.rst
+++ b/Documentation/DataTypes/Align/Index.rst
@@ -7,6 +7,11 @@
 align
 =====
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


Chrissitopher:
It is currently documented that this data type is used at many places inside the core; not only in IMGTEXT and CTABLE. 

For the documentation, we could "fix" this by writing "string" as data type where we currently say "align". However, this would not change the code behind all those different properties. Or are there deprecation notices telling that all the according properties would be deprecated?

marble:
Hi Elmar, thanks for pointing to the issue. This kind of "data-type" still exist and will keep on existing. It only has the meaning that something needs to be aligned. By whatever method. But HTML shouldn't be mentioned, indeed.

Issue #48, opened 2015-11-12
Fix includeLibs example


Chrissitopher:
(This is a follow-up to #38.)

In Functions/Stdwrap/Index.rst, 
https://docs.typo3.org/typo3cms/TyposcriptReference/Functions/Stdwrap/Index.html#postuserfunc
there is one line with 

``includeLibs.reverseString = fileadmin/example_reverseString.php``

Problem is, since TYPO3 7, the property includeLibs has been removed (forge #67646) - that is the reason why the example now is no longer working.

According to the migration notes, now a hook should be used during the Frontend set up to execute custom PHP code?

How can the example be updated to be working again?

Chrissitopher:
Fixed with #51.

PR #51, opened 2015-11-22
[BUGFIX] Align userFunc example with current code

wmdbMattes:
We no longer allow includeLibs.
Furthermore the example has been adapted to utilize namespaces.


wmdbMattes:
Fixed #48

Chrissitopher:
Thanks for the fix, Mattes!

helhum:
Thanks @wmdbMattes !

PR #53, opened 2015-12-08
How to check empty or not empty

Gernott:
Check, if a value is empty or is not empty with regular expression.


marble:
Very good, thank you very much, and: Give me more !-) BTW, you need to write '::' to make the following blank line plus indented lines a code block. In the TypoScript Reference the default highlighting is "typoscript", so this fits. I've reworked the page a bit and included some more headlines in the examples, mainly because at the moment the theme doesn't have a good "bottom margin" at the code-blocks. Reload your browser: https://docs.typo3.org/typo3cms/TyposcriptReference/Conditions/Index.html#test-for-empty-value

PR #59, opened 2016-04-06
Automatic visibility control of records

elmar-hinz:
I think it useful to indicate the influence of the select query to hide some records automatically. It's basic information of the overall usefulness of this content element.


marble:
Thank you Elmar. I've rephrased the words a bit, and once I was at it, reworked some reST to better practises.

PR #60, opened 2016-04-18
 Updated examples not to use fileadmin

sgrossberndt:
The examples for includung CSS and JavaScript files now point to EXT:extension/Resources/Public/ instead of fileadmin since we want to educate people to use extensions to provide files instead of putting them into fileadmin.


marble:
What about the wildcard ...template*.css that was in the original example? Did you drop that on purpose?

PR #68, opened 2016-07-07
[Task] add default for pageRendererTemplateFile

erredeco:
for better understanding of config.pageRendererTemplateFile I add its default value;
related issue: https://forge.typo3.org/issues/60723


marble:
Very good - thanks!

PR #69, opened 2016-07-14
Update Index.rst

Ichag:
Make TypoScript example for Layouts more consistent and add an example for Partials.


marble:
Thanks for spotting these inconsistencies. Hower, I'm not sure your patch is totally correct. I've made some little changes. Do you think that's ok? 
https://docs.typo3.org/typo3cms/TyposcriptReference/6.2/ContentObjects/Fluidtemplate/Index.html
Besides that I updated the reST code a bit (one tab -> three blanks, headlines ...)

Ichag:
If it makes it better then before I'm totally green with the changes :+1:

PR #70, opened 2016-07-26
fix 1 one-letterTypo

knbknb:
the Property name is shown as 
  Property
         defaultGetVars

but the code example has it in weird mixed lowercase/uppercase mix


marble:
Thank you. Don't hesitate to send more.

PR #73, opened 2016-09-05
update example

macjohnny:
disable translation handling of sys_category, otherwise only translated records of categories are retrieved instead of all actually selected.


macjohnny:
backport to 7.6

marble:
Applied to: latest, 7.6

PR #75, opened 2016-09-11
Add link to documentation of ext:form

tritumBj:


marble:
Hi Björn, it's merged. I made a mistake so Github doesn't show it as merged. Once I was at the file I did quite a bit of reworking it. Looks nicer now here and there. See https://docs.typo3.org/typo3cms/TyposcriptReference/ContentObjects/Form/Index.html and https://docs.typo3.org/typo3cms/TyposcriptReference/ContentObjects/Form/Index.html#example-mailform

tritumBj:
Hi Martin. Thanks a lot :) You're version is even better since it's clearer. Thx again.

PR #76, opened 2016-09-30
prefixLocalAnchors requires compatibility6 extension

thegass:
add hint to the mandatory compatibility6 extension if prefixLocalAnchors is required.


Chrissitopher:
This property should no longer be documented there at all.

PR #80, opened 2016-11-23
make reference to addRootLineFields more specific

tmotyl:


tmotyl:
is the addRootLineFields documented somewhere? I tried to find a crosslink but couldn't

marble:
Too bad, no, it isn't: https://www.google.de/search?q=site%3Adocs.typo3.org+addrootlinefields Only mentioned in extension manuals.

PR #82, opened 2016-11-24
Update Index.rst (optionsplit example)

pi-phi:
new column with number of items to explain different results for optionsplit

### DIFF
--- a/Documentation/ObjectsAndProperties/Index.rst
+++ b/Documentation/ObjectsAndProperties/Index.rst
@@ -381,17 +381,17 @@ Overview with abstract examples
 The following table gives you a condensed overview of how the
 optionSplit rules work together:
 
-=========================  ======================  =======
-optionSplit                Resulting items         Rule
-=========================  ======================  =======
-a                          a a a a
-a || b || c                a b c c c ...
-a || b \|*| c              a b c c c ...
-a || b \|*| c \|*| d || e  a b c c ... c c d e     Rule 1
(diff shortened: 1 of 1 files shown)


Chrissitopher:
What should these numbers be good for? As far as I understand it, the number of items in the end is determined by - if it's a menu - the number of pages in that menu. For the first example with only "a": If the menu only has e.g. two pages, then there will only be two items, not four and so on for the other examples.

pi-phi:
The numbers should explain the example results as these result depend on the number of items. Especially the examples 4,5 and 6 have the same optionsplit. For a more systematic approach it would be helpful to see results of each optionsplit for 2,3,4,5,6,7 elements.

marble:
Hi Bernd, hooray. You are the one and only hero up to now who has dared to touch that page. It is really still in a miserable condition. Almost directly what we got after automatic conversion of the old OpenOffice documents.
I have picked up the idea of your PR but didn't use it directly. Please see https://docs.typo3.org/typo3cms/TyposcriptReference/ObjectsAndProperties/Index.html#overview-with-abstract-examples  For each line of the old example table (which is still there but should be dropped) I have created a new table which should show the result sequence for different numbers of input elements.
*Note:* I wrote the result lines by heart and without testing. Could you do a review on whether the tables are correct?
Feel free to change things or add stuff.

marble:
I'm closing this PR because it has become "obsolete". What I mean: We have started communication on Slack and by mail. And as a consequence the whole page is going to be reworked. Let's wait for this.

PR #88, opened 2016-12-29
Update Index.rst

pi-phi:
some minor fixes: typo, code cleaning


marble:
Bernd, would you like me to give credit to you somewhere? Maybe in the news on the docs.typo3.org homepage? I wasn't sure you would want it. It would be less than fair!

pi-phi:
kein Problem, aber die Leute sehen das doch sicher auch in der History dass ich da mitgeholfen habe

PR #89, opened 2017-01-03
Wrong Propertyname

simare:
Hello,
shouldn't it read '.templateName' instead of '.template' ?


marble:
Can you find out the answer and then come back here?

simare:
This was more of a rhetorical question, it really should read 'templateName'.
But the person merging this should know if it is correct or not, in my opinion.

marble:
I'm merging it but unfortunately can't know everything. You are invited to join the doc team and help merging :-))

PR #102, opened 2017-06-12
Change default value of removeDefaultJs

manuelselbach:
As mentioned in the `TypoScriptFrontendController.php` (line 2522) the default value would be `external` instead of `1`

```
// Set default values for removeDefaultJS and inlineStyle2TempFile so CSS and JS are externalized if compatversion is higher than 4.0
if (!isset($this->config['config']['removeDefaultJS'])) {
    $this->config['config']['removeDefaultJS'] = 'external';
}
```

### DIFF
--- a/Documentation/Setup/Config/Index.rst
+++ b/Documentation/Setup/Config/Index.rst
@@ -2282,7 +2282,7 @@ removeDefaultJS
             config.removeDefaultJS = 1
 
    Default
-         1
+         external
 
 
 


Chrissitopher:
Merged with https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/commit/62d3ac8301b7e8cc6d57e3ab1518d890810678c0.

PR #103, opened 2017-06-13
Add another example to cssInline

simonschaufi:
Add another example to cssInline and remove a line break which is not correct typoscript syntax


Chrissitopher:
Thanks for the PR, Simon!

PR #104, opened 2017-06-28
FIX: substring desc mentions wrong php function

Kitzberger:


marble:
Great. Thanks. Send more!

PR #105, opened 2017-07-06
Update description of the language related settings

tmotyl:
Make description of the sys_language_uid, sys_language_overlay and sys_language_mode more verbose and precise.
See typo3/sysext/frontend/Tests/Functional/Rendering/LocalizedContentRenderingTest.php
for a test covering different configuration combinations.

### DIFF
--- a/Documentation/Setup/Config/Index.rst
+++ b/Documentation/Setup/Config/Index.rst
@@ -2557,33 +2557,46 @@ sys\_language\_mode
          string
 
    Description
-         Setting various modes of handling localization.
-
+         Configures what system should do when a page is not translated
+         to the requested language.
+         It is only evaluated when `sys_language_uid` is greater than `0`,
+         and the requested page translation is NOT available.
+         Internally this setting influences the value set to the 
+         `$GLOBALS['TSFE']->sys_language_content`.
+         
          The syntax is "[keyword] ; [value]".
 
-         Possible keywords are:
+         **Possible keywords are:**
 
-         [default] - The system will look for a translation of the page (from
-         "Alternative Page Language" table) and if it is not found it will fall
-         back to the default language and display that.
+         [empty] - If not set, and the page is not translated, the system will
+         behave as if the default language was requested.
+         Internally both `$GLOBALS['TSFE']->sys_language_content`
+         and `$GLOBALS['TSFE']->sys_language_uid` whill be set to 0;
 
-         **content\_fallback:** Recommended. The system will always operate
+         **content\_fallback** - Recommended. The system will always operate
          with the selected language even if the page is not translated with a
          page overlay record. This will keep menus etc. translated. However,
          the *content* on the page can still fall back to another language,
-         defined by the value of this keyword, e.g. "content\_fallback ; 1,0"
+         defined by the value of this keyword, e.g. :ts:`content_fallback;1,3,0`
          to fall back to the content of sys\_language\_uid 1 and if that is not
          present either, to default (0).
+         Note that the fallback is on the page level, not on the content level.
+         This means that once a translated page record is found in the fallback
+         chain, system will try to use this language for the rendering even
+         if there is no content translated.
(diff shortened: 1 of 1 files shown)


marble:
@tmotyl Good work! cc/ @andreaswolf 

And, Andreas, I saw you comments only after making my updates. Maybe I got most of what you're adding already. Everybody is welcome to continue improving.

IMHO: If something in the text is questionable and we are not really sure right now we should leave note in the text somehow. Maybe like '((this is the proposed behaviour - needs to be confirmed))'.

Documentation/Setup/Config/Index.rst
```diff
@@ -2557,33 +2557,46 @@ sys\_language\_mode
...
-         Possible keywords are:
+         **Possible keywords are:**
 
-         [default] - The system will look for a translation of the page (from
-         "Alternative Page Language" table) and if it is not found it will fall
-         back to the default language and display that.
+         [empty] - If not set, and the page is not translated, the system will
+         behave as if the default language was requested.
+         Internally both `$GLOBALS['TSFE']->sys_language_content`
+         and `$GLOBALS['TSFE']->sys_language_uid` whill be set to 0;
```
tmotyl: only when the page translation is not available. E.g. you request L=3, there is no page overlay with for lang 3, so 
`$GLOBALS['TSFE']->sys_language_content` and `$GLOBALS['TSFE']->sys_language_uid` will both be set to 0.
Also see the note at the beginnig => "It is only evaluated when :ts:`sys_language_uid` is greater than `0` and the requested page translation is NOT available."

Documentation/Setup/Config/Index.rst
```diff
@@ -2557,33 +2557,46 @@ sys\_language\_mode
...
          page overlay record. This will keep menus etc. translated. However,
          the *content* on the page can still fall back to another language,
-         defined by the value of this keyword, e.g. "content\_fallback ; 1,0"
+         defined by the value of this keyword, e.g. :ts:`content_fallback;1,3,0`
          to fall back to the content of sys\_language\_uid 1 and if that is not
          present either, to default (0).
+         Note that the fallback is on the page level, not on the content level.
+         This means that once a translated page record is found in the fallback
+         chain, system will try to use this language for the rendering even
+         if there is no content translated.
```
tmotyl: ah, you;re right :(  I've just tested it again and right now the fallback chain is not changing the page record. I consider this a bug see (https://review.typo3.org/#/c/53307/), but nevertheless for now this sentence has to go away from the docs.

Documentation/Setup/Config/Index.rst
```diff
@@ -2557,36 +2546,57 @@ sys\_language\_mode
...
-
-         An in-depth discussion is found in the
-         :ref:`Frontend Localization Guide <t3l10n:localization-modes>`.
+         **Possible keywords are:**
+
+         [empty] (not set)
+            If not set and the page is not translated, the system will
+            behave as if the default language was requested.
+            Internally both :php:`$GLOBALS['TSFE']->sys_language_content`
+            and :phpp:`$GLOBALS['TSFE']->sys_language_uid` whill be set to `0`.
```
mbrodala: Here's a typo. ;-)

Documentation/Setup/Config/Index.rst
```diff
@@ -2557,36 +2546,57 @@ sys\_language\_mode
...
+
+         content\_fallback
+            Recommended. The system will always operate
+            with the selected language even if the page is not translated and has no
+            page overlay record. This will keep menus etc. translated. However,
+            the *content* on the page can still fall back to another language,
+            defined by the value of this keyword, e.g. :ts:`content_fallback;1,3,0`
+            to fall back to the content of sys\_language\_uid 1 and if that is not
+            present either, to default (0).
```
mbrodala: There is no mention of `3` anymore here which will be confusing.

PR #109, opened 2017-08-23
Use proper example for stdWrap's date

Kitzberger:


marble:
Thank you. Send more!!!

PR #111, opened 2017-09-27
Update stdWrap_bytes

megamisan:
Make the documentation match actual implementation changed 2 years ago during June 2015 sprint.

### DIFF
--- a/Documentation/Functions/Stdwrap/Index.rst
+++ b/Documentation/Functions/Stdwrap/Index.rst
@@ -1023,34 +1023,51 @@ bytes
    Data type
          integer /stdWrap
 
+   Default
+         iec, 1024
+
    Description
-         Will format the input (an integer) as bytes: bytes, kb, mb, …
+         This function formats an integer value. The output will be one of the
+         usual byte notations like "bytes", "kb", "mb" and so on. The exact
+         behavior is defined by sub-properties. Compare with
+         `Wikipedia "Kibibyte" <https://en.wikipedia.org/wiki/Kibibyte>`__
+         for example.
 
-         **Available sub-properties:**
+         .labels
+            With :ts:`.labels = iec` the common units of the IEC are used
+            which have a base of 1024. Other values of base will be ignored.
 
-         **labels**:
-         If you add a value for the property "labels" you can alter the default
-         suffixes. Accept a preset keyword or labels for bytes, kilo, mega, giga
-         and bigger. Labels are separated by vertical bar (\|) and possibly encapsulated
-         in "". E.g.: " \| K\| M\| G". The default is the keyword `iec`
-         which is backward-compatible.
-         
-         **base**:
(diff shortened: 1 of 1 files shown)


Documentation/Functions/Stdwrap/Index.rst
```diff
@@ -1024,22 +1024,25 @@ bytes
...
 
          **Available sub-properties:**
 
          **labels**:
          If you add a value for the property "labels" you can alter the default
-         suffixes. Labels for bytes, kilo, mega and giga are separated by
-         vertical bar (\|) and possibly encapsulated in "". E.g.: " \| K\| M\| G"
-         (which is the default value).
-
+         suffixes. Accept a preset keyword or labels for bytes, kilo, mega, giga
```
marble: I don't understand this sentence (?): "Accept ..." What do you mean?

megamisan: The field accept either of a preset keyword or a list of label. Should I try reformulate? Any suggestion?

marble: Sure! You should exactly have written that :-) (The field accepts either ...) It's ok, you don't have to send an update. I'll make it fit and check it in tomorrow.

megamisan:
Shouldn't we also patch documentation for 8.7 LTS ?

marble:
Ah, the more I worked on this, the longer it took In the end I've now reworked that whole chapter 'bytes'. Can you please read through it and check that I did it right? Because I didn't test that and actually never used that function. But it's quite useful, I think. Once it's ok we should forward it to the 'latest' branch as well.

marble:
You got it. You were too fast :-))

marble:
Please wait 5 minutes or so until it's online https://docs.typo3.org/typo3cms/TyposcriptReference/Functions/Stdwrap/Index.html#bytes

marble:
Ok, there it is. https://docs.typo3.org/typo3cms/TyposcriptReference/7.6/Functions/Stdwrap/Index.html#bytes
Please check. Do you have some examples of the output at hand? Would be great to add those.

megamisan:
Well, it looks good, but as you can see, the backslash in the label string example in "labels" property are extraneous.

Examples:
* si:
** input=8456123 => output=8.45 M
** input=5515312101012315531 => output=5.51 E
* iec:
** input=8456123 => output=8.06 Mi
** input=5515312101012315531 => output=4.78 Ei
* custom:
** input=8456123 => output=8.06 mebibyte
** input=5515312101012315531 => output=5136534665.72 gibibyte

marble:
I've finished this now. Unbelievable, how much time it takes, to get a single paragraph right.
- https://docs.typo3.org/typo3cms/TyposcriptReference/Functions/Stdwrap/#bytes
- https://docs.typo3.org/typo3cms/TyposcriptReference/7.6/Functions/Stdwrap/Index.html#bytes

PR #114, opened 2017-11-08
Update Index.rst, make working example for "variables" property of FLUIDTEMPLATE

hg2000:
In the example using "variables.foo = bar" seems irritating to me. The user expects to be able to call "foo" in the template then via {foo}. But that does not work because you have to use a CObject.


marble:
Thank you. Send more!

PR #115, opened 2017-11-13
add information about format

macjohnny:
@marble is there anything more we could add about the FLUIDTEMPLATE format property?
where is it used? what can we do with it?
I just copied the text found at 
https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/extbase/Classes/Mvc/Request.php#L465

From the path resolution in 
https://github.com/TYPO3/Fluid/blob/master/src/View/TemplatePaths.php#L253
it looks like this is the file suffix for the template file that will be used.

### DIFF
--- a/Documentation/ContentObjects/Fluidtemplate/Index.rst
+++ b/Documentation/ContentObjects/Fluidtemplate/Index.rst
@@ -302,6 +302,7 @@ format
 
    Description
          Sets the format of the current request.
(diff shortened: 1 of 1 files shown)


marble:
Thank you. Ok with you?

macjohnny:
@marble how about adding an example how to use it?

marble:
Great. I love examples. Can you create one or some? Just give it a try, I can do the finish if you want.

PR #116, opened 2017-11-27
#54788 Cache-control header value change

lukaszuznanski:
Change in Cache-control header value related to https://review.typo3.org/#/c/54788/


liayn:
Has to wait for: https://review.typo3.org/54788

PR #117, opened 2017-12-02
change position of default block

erredeco:
"default" values block is rendered in the wrong position; plus: streamline of default boolean values to "0 (false)"


marble:
Cool, you got it. Feel free to move 'Default' on other pages as well.

PR #121, opened 2017-12-03
Use "Table" layout for properties

erredeco:
Modified layout for properties to make them consistent with the other parts of documentation; please review and check the title formatting (I am not sure that they are correct)


marble:
you need to add the "table comment" lines to make this ready for a merge. Just push that to you patch-4 branch and it will appear here.

marble:
You'll have to practise this a bit it seems :-)) Let's see if I got it right.

PR #130, opened 2017-12-04
change position of default block

erredeco:
"default" values block is rendered in the wrong position;
[NOTE] I don't know if it sufficient to solve this issue: the title of section "extbase.pluginName" floats (see https://docs.typo3.org/typo3cms/TyposcriptReference/ContentObjects/Fluidtemplate/Index.html#extbase-pluginname)


marble:
let's try

PR #135, opened 2018-02-12
Add missing value and description

mieserfettsack:
As of TYPO3 8.7 you need to enable rounding with round = 1

https://forge.typo3.org/issues/83859


marble:
Shall we backport this to 8.7?

marble:
Backported to 8.7!

mieserfettsack:
Porting to 8.7 was right! Thx!

PR #137, opened 2018-02-15
TAKS: Add info how to access labels in JS

DanielSiepmann:


marble:
Please check whether my update is ok with you.

PR #139, opened 2018-03-13
Add example for if.equals

christophlehmann:


marble:
Thank you. Send more!

PR #148, opened 2018-04-08
[TASK] Remove the mention of not longer used strlen function

gilbertsoft:
With the Task #54091 strlen compares to zero are replaced by a check against '' (empty string).


froemken:
This property is handled that way since years...maybe since it was added to TYPO3. So, yes, it would be good to have a backport to 8, 7, 6, 4, ...

PR #149, opened 2018-04-12
Optimize example for DataProcessing\SplitProcessor

julianhofmann:
Make example work. Filtering intergers in processor, but outputting keywords in template contradicts itself.


marble:
Thank you. It looks like you know what you are doing. Send more!

PR #158, opened 2018-06-14
correction of fancybox example

fabtho:


marble:
Thank you. Send more!

PR #160, opened 2018-07-05
write it in the positive way

franzholz:
Thinking positively is easier than thinking negatively.


franzholz:
The meaning does not change.
You have had sentences like this:

isGreaterThan:

returns false if the first value is not greater than the second value

I have change this into:

returns true if the first value is greater than the second value.

a > b ? 
false : a <= b
true: a > b


Both explanations have exactly the same meaning.

DanielSiepmann:
Yes the meaning is the same. But it's much easier to read and understand if it's not `false` and `not`. Thanks.

PR #161, opened 2018-07-23
Update Constants.rst

Rugguduh:
fix (suspected) typo in line 80: "finding" instead of "fining"


marble:
Thank you. Send more!

PR #163, opened 2018-08-02
Correct the <picture> descriptions

srosskopf:
There's a mistake in the `<picture>` element related sections. The `<source>` tags inside `<picture>` need to have a `srcset="..."` image set instead of `src="..."`. Otherwise the picture layout will not work.

See https://www.w3schools.com/tags/tag_source.asp for more details.


marble:
Looks like you know what you're doing. Great! Thanks, and: Send more!

PR #164, opened 2018-08-30
add link to ContentObjectRenderer.php

franzholz:


marble:
A while ago I added many links to the API in various manuals. Only to find out that the API changed quickly and those links became invalid. Moreover: I know of considerations to drop the api site completely. So it seems it doesn't make sense to add those kind of links. I therefore just close this PR. Though in general I would like the idea. Thanks for trying!

PR #178, opened 2018-09-05
Extract GMENUITEM from GMENU->

franzholz:
Additional properties for Menu item states
should come into GMENU


franzholz:
Some references will be adapted later when this is in.

PR #181, opened 2018-09-13
Add documentation for the new TypoScript condition

janhelke:
Thanks to Frank who provided an awesome .rst file.


marble:
Ok, this looks like a bigger change. Thanks for contributing. I've assigned myself to the PR. Before merging I'll investigate whether we should include some information that helps to find the <9.4 docs. Please gimme a moment.

PR #184, opened 2018-09-25
The former sentence had no meaning.

franzholz:


marble:
That sentence sounds much better. Tip: Write "imperative" commit message, like a command. The idea is, that people know what they will be doing when they apply that commit. IIRC If I remember correctly that's well described here.

Thanks for your work!

PR #189, opened 2018-11-22
Add info about the deprecated cObject FILE

josefglatz:


sypets:
@josefglatz Might be nice to add a link to the Changelog because it contains info about migration. You mean: https://docs.typo3.org/typo3cms/extensions/core/latest/Changelog/9.5/Deprecation-85970-FileContentObject.html ?

I will push another commit.

PR #190, opened 2018-11-24
reference to function name

franzholz:


franzholz:
The function name should link to:

https://docs.typo3.org/typo3cms/TyposcriptReference/DataTypes/Index.html#function-name

Is something still missing?

sypets:
@franzholz Ah, thanks. Reference should be `:ref:`data-type-function-name``. I pushed another commit to fix and will merge. Thanks.

PR #193, opened 2018-12-28
Update Index.rst

janhelke:
Add configuration for new data type siteLanguage. Introduced in https://review.typo3.org/#/c/58930/

### DIFF
--- a/Documentation/DataTypes/Index.rst
+++ b/Documentation/DataTypes/Index.rst
@@ -645,7 +645,7 @@ rotation
 .. _data-type-siteLanguage:
 
 siteLanguage
-----
+------------
 
 :aspect:`Data type:`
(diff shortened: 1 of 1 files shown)


Documentation/DataTypes/Index.rst
```diff
@@ -642,7 +642,45 @@ rotation
 :aspect:`Example:`
    180
 
-.. _data-type-space:
+.. _data-type-siteLanguage:
+
+siteLanguage
+----
```
sypets: Section title underlines should be at least as long as the title. By convention used in TYPO3 docs, they should be exactly as long as title, see https://docs.typo3.org/typo3cms/HowToDocument/WritingReST/HeadlinesAndSection.html#syntax

PR #194, opened 2019-01-07
A .ico file is no longer required for page.shortcutIcon

sorenmalling:
The icon used for page.shortcutIcon is no longer required to be a .ico file

This commit remove the *note* part that mentions that


DanielSiepmann:
Thanks for updating this part @sorenmalling.

PR #196, opened 2019-01-21
add description for lang-key

franzholz:


DanielSiepmann:
Thanks @franzholz for adding further helpful explanations.

PR #197, opened 2019-01-22
Fix typoscript condition

Smove:
As commented in AbstractConditionMatcher.php (and seen on line 676 in the same rst) the operator has to be `>=` instead of `=>`


DanielSiepmann:
Nice catch @Smove, thanks for the fix.

PR #199, opened 2019-03-06
clarification levelfield

pi-phi:
Indication of the need to declare rootline fields in Install Tool,
adding `.data` to get proper content for property 'foo'

### DIFF
--- a/Documentation/DataTypes/Properties/GetText.rst.txt
+++ b/Documentation/DataTypes/Properties/GetText.rst.txt
@@ -392,6 +392,7 @@ levelfield
 :aspect:`Description:`
    levelfield: Like :ref:`data-type-gettext-leveltitle` et al. but where the
    second parameter is the root line field to fetch.
+   If you use other fields than the usual be sure to add them in the InstallTool to the rootline fields. ( [FE][addRootLineFields] )
 
 :aspect:`Example:`
    Get the value of the user defined field user_myExtField in the root line.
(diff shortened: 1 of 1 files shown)


Documentation/DataTypes/Properties/GetText.rst.txt
```diff
@@ -400,7 +401,7 @@ levelfield
    field. In order that you can use this function, your fieldname 'user_myExtField' needs be included in the comma
    separated list of ['addRootLineFields']::
 
-      foo = levelfield : -1, user_myExtField, slide
+      foo.data = levelfield : -1, user_myExtField, slide
```
DanielSiepmann: I'm not sure about adding the `.data`. It's nowhere and I guess it should not be added as one could use `insertData` and further syntax. This is about the datatype itself, not how to use it in certain circumstances.

What do you think @pi-phi?

sypets:
This PR was automatically closed by GitHub because of housekeeping tasks we had to do. The "latest" branch is no longer active, we now use "master". (see [2019-06-01 - [IMPORTANT] Change Main Branches to master] (https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/TipOfTheDay.html#tip-branches-master). 

Additionally, there were open questions and no activity for some time.

If you feel your change is still important, please create a new pull request or contact us in Slack.

Thank you for your contributions.

PR #200, opened 2019-04-01
Update location of optionSplit PHP function

mbonaker:


DanielSiepmann:
Thanks @mbonaker for updating the outdated class name.

PR #201, opened 2019-04-05
Escape * as it is needed for function like

froemken:
I don't know, if I have escaped it correct for Sphinx. Please update, if needed.

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -739,7 +739,7 @@ like
 :aspect:`Example`
    Search a string with ``*`` within another string::
 
-      [like("fooBarBaz", "*Bar*")]
+      [like("fooBarBaz", "\*Bar\*")]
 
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
So the `*` would not work without being escaped? Is only for rendering due to RST / Sphinx, or due to handling / parsing of the string within TYPO3 / Symfony?

I've pushed an update to reflect changes to the manual, in order to keep your pull request mergeable.

froemken:
Hmmm...in my documentation I have never used escaping of `*` in `like`: https://www.typo3lexikon.de/typo3-tutorials/typoscript/referenz/condition-syntax.html
I have just tested it again and it works without escaping, so I'm closing this issue.

DanielSiepmann:
Thanks for investing the time @froemken.

Btw, your reference is now integrated into TYPO3 documentation: https://docs.typo3.org/typo3cms/TyposcriptReference/Conditions/Index.html#reference

PR #205, opened 2019-04-15
Fix order and association of headlines to examples

moe2k:
Fix order and association of headlines to examples in "pageTitleSeparator" section.

### DIFF
--- a/Documentation/Setup/Config/Index.rst
+++ b/Documentation/Setup/Config/Index.rst
@@ -1966,22 +1966,20 @@ pageTitleSeparator
          the end of the separator.
 
    Examples
-         ::
+
+         This produces a title tag with the content "website . page title"::
 
             config.pageTitleSeparator = .
 
-         This produces a title tag with the content "website. page title"::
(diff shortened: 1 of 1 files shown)


Documentation/Setup/Config/Index.rst
```diff
@@ -1968,21 +1968,20 @@ pageTitleSeparator
    Examples
          ::
```
sypets: This extra `::` should be removed.

Documentation/Setup/Config/Index.rst
```diff
@@ -1968,21 +1968,20 @@ pageTitleSeparator
...
 
+         This produces a title tag with the content "website. page title"::
```
sypets: On my end, this actually results in "website . page title" (space before and after ".").

DanielSiepmann: Right, due do not using `config.pageTitleSeparator.noTrimWrap = |||` in this example.

DanielSiepmann:
Thanks @moe2k for providing bug fixed to wrong documentation.

PR #209, opened 2019-07-02
Add example for getTSFE() with current typeNum

sebkln:


DanielSiepmann:
Thanks for your first contribution. Formatting was fine, I've just added a cross reference to typeNum property.

PR #217, opened 2019-08-16
[TASK] Update with changes from 9.5

susannemoog:
* [ ] master
* [ ] 9.5


Documentation/Setup/Config/Index.rst
```diff
@@ -1259,7 +1264,57 @@ htmlTag\_setParams
...
+
+   Data type
+         :ref:`data-type-array`
+
+   Description
+         Sets the attributes for the :html:`<html>` tag on the page. Allows to override 
+         and add custom attributes via TypoScript without having to re-add the 
+         existing attributes generated by SiteHandling.
+
+         This property supersedes the previous `config.htmlTag_setParams` option by providing
```
Lefaux: does it make sense to write this at htmlTag_setParams as well?

susannemoog: htmlTag_setParams got these two infos:


.. warning::
    If you are using `htmlTag.atttributes` this property (`htmlTag_setParams`) will not have any effect.

.. note::
    Used to older TYPO3 versions? If you are using site configuration you do not need to set `htmlTag_setParams` for language 
    related configuration in TypoScript - use the site configuration instead.

PR #222, opened 2019-08-28
[BUGFIX] Header level

DanielSiepmann:
site and siteLanguage is sub property of get text and should therefore
be on same level as all other sub properties.

Releases: 9.5, master


DanielSiepmann:
For 9.5: #221

PR #238, opened 2019-09-27
Update Menu object common properties

tmotyl:
To be compatible with descriptions from /ContentObjects/Hmenu/Index.rst
Add deprecation warnings see
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.4/Deprecation-85902-IMGMENUGMENU.html


tmotyl:
Thank you!

Issue #237, opened 2019-09-27
Content objects are missing cache property


tmotyl:
since https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/7.4/Feature-64200-AllowIndividualContentCaching.html

Every content object supports `cache` property.

DanielSiepmann:
This is not yet included into all cObjects, due to different markup usage. See #239 for further information.

linawolf:
@DanielSiepmann @tmotyl can one of you make a PR about that?

tmotyl:
@linawolf sorry, in the next 2months I would not be able to dig into that, I'm totally green in rst markup stuff and what is considered best practice in docs (why some pages are done in one way and others are different)

DavidBruchmann:
@tmotyl don't mind about rst syntax, if you could add the important info just in text format that would be absolutely fine, the documentation team will care about the formatting.

tmotyl:
@DavidBruchmann its already documented in some cobjects , see change: 
https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/pull/239/files
This ticket is about adding the same information to other objects too.

DavidBruchmann:
@tmotyl thanks for the clarification, that makes it easier to work on it.

PR #239, opened 2019-09-27
Include Cache property to cObjects

tmotyl:
solves https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/issues/237

### DIFF
--- a/Documentation/ContentObjects/Case/Index.rst
+++ b/Documentation/ContentObjects/Case/Index.rst
@@ -115,6 +115,7 @@ from the "default" property will be used.
          stdWrap around any object that was rendered no matter what the "key"
          value is.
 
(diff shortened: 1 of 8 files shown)


tmotyl:
not sure whether the approach I took is ok and how to proceed.
Basically all ContentObjects shoudl get the cache property

tmotyl:
Tested it with local docker rendering chain, looks good

DanielSiepmann:
It is intended not to add the property to all cObjects? E.g. FLUIDTEMPLATE did not receive an update.

tmotyl:
It should be added everywhere, but some content objects have different table structure than others so I couldnt just include the same file.

PR #240, opened 2019-10-03
Propose a more understandable name for a page object

tmotyl:
Open for a discussion how to make it more verbose from the menu and manual content what this property is about.


mbrodala:
How about something like `page, ...`? Most of the time only `page` exists though, so that should be enough.

tmotyl:
ok for me :)

PR #242, opened 2019-10-18
JavaScript script tags omit type=text/javascript in HTML5

brotkrueml:
See: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.0/Breaking-88772-JavaScriptScriptTagsOmitTypetextjavascriptInHTML5.html


mbrodala:
Thanks. :-)

PR #251, opened 2019-11-11
Improve rendering of code example for 'PageTitleProviders'

sebkln:
The examples are not rendered inside the container 't3-row'
because of an unnecessary blank line.

Releases: master

### DIFF
--- a/Documentation/Setup/Config/Index.rst
+++ b/Documentation/Setup/Config/Index.rst
@@ -2024,7 +2024,6 @@ pageTitleProviders
          the lowest in priority.
 
    Examples
-
          By default, TYPO3 ships with two providers::
 
             config.pageTitleProviders {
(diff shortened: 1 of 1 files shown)


sebkln:
I created a second PR for 'pageTitleSeparator' because of different release targets. The second needs to be fixed in 9.5, too.

I thought it would be easier to merge into the different version branches this way. Isn't it?

sypets:
Ah sorry, I already pushed. I will take care of it and the backport too. 

You are right, if the older version differs it may be easier to backport. (Now I have 2 commits and I can cherry-pick only one commit 2 the other branch, so should not be a problem.).

sebkln:
Okay. Originally I had one PR for all changes in _config_ (instead of the three) prepared, but then noticed the differences in the version branches.

I know it's the most simple PRs you could create, but you'll have to start somewhere. ;)

sypets:
It's very much appreciated. :thumbsup: Also great that you are thinking about the other branches as well.

PR #253, opened 2019-11-15
Clarify inclusion of JavaScript

spoonerWeb:
* Set block "same as includeJS" to section "includeJSFooter" as well
* Add short descriptions of what the difference is between all "includeJSxxx" sections
* Move example in "includeJSFooter" to "includeJS"

### DIFF
--- a/Documentation/Setup/Page/Index.rst
+++ b/Documentation/Setup/Page/Index.rst
@@ -533,6 +533,10 @@ includeJS.[array]
                 # in the TS constants field.
                 file2 = javascript_uploaded_to_template*.js
                 file2.if.isTrue = {$myConstant}
+                
+                twitter = http://twitter.com/javascripts/blogger.js
+                twitter.external = 1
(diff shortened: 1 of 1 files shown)


Documentation/Setup/Page/Index.rst
```diff
@@ -533,6 +533,10 @@ includeJS.[array]
                 # in the TS constants field.
                 file2 = javascript_uploaded_to_template*.js
                 file2.if.isTrue = {$myConstant}
+                
+                twitter = http://twitter.com/javascripts/blogger.js
```
mbrodala: This does not exist. And if, it should use HTTPS and probably a fixed version to ensure the mentioned `integrity` matches.

spoonerWeb: This was just moved from the doubled section I removed. Didn't check if this JS exists. What about using another JS like the latest of jQuery?

mbrodala: Good idea. :-)

spoonerWeb: Using jQuery now.

PR #274, opened 2020-01-10
give a hint about a link without html a tags

franzholz:
It is very hard to find out, how to get a link and have not a tags around it.


sypets:
Thank you :thumbsup: 

While looking at the page, I noticed some other issues, see #275 and #276.

sypets:
I backported the change to the older version.

PR #286, opened 2020-02-26
Change misleading description for inlineSettings

IndyIndyIndy:
extJS is no longer part of TYPO3. Change the misleading description text for inlineSettings.


sypets:
Cherry-picked to 9.5, 8.7

PR #287, opened 2020-02-26
Update list of excludeDoktypes

erredeco:
The list of `excludeDoktypes` still included the long-time removed value "5"  -see also https://forge.typo3.org/issues/90472

Please review and backport at will. Thank you.


sypets:
Thank you :thumbsup:

Backported to all branches.

PR #291, opened 2020-03-07
Get rid of wrong quotation marks

funkvn:


sypets:
cherry-picked to newer versions.

PR #292, opened 2020-03-10
[BUGFIX] Adjust changelog-URLs to current URL-structure

neufeind:


sypets:
Thanks for the patch. But we can use the newly proposed formatting, for example:

    :doc:`t3core:Changelog/8.1/Deprecation-75625-DeprecatedCacheClearingOptions`

This way, the base URL is maintained in Settings.cfg and does not have to be changed in several places. 

see:

https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/HowToUpdateDocs.html#link-to-changelog

I can take care of this. In the future, please use the new formatting. I will also add "t3core" to the Settings.cfg in the main repos shortly.

neufeind:
Alright, I wasn't aware of the new format - just noticed the URL didn't work anymore in the rendered documentation.
It would be great if you could take care, also for the various branches. Thanks.

PR #294, opened 2020-03-21
cObjGetSingle can have a third parameter

franzholz:
/**
     * Renders a content object
     *
     * @param string $name The content object name, eg. "TEXT" or "USER" or "IMAGE
     * @param array $conf The array with TypoScript properties for the content object
     * @param string $TSkey A string label used for the internal debugging tracking.
     * @return string cObject output
     * @throws \UnexpectedValueException
     */
    public function cObjGetSingle($name, $conf, $TSkey = '__')
    {

### DIFF
--- a/Documentation/AppendixA/Index.rst
+++ b/Documentation/AppendixA/Index.rst
@@ -84,10 +84,10 @@ Example:
 
 .. _appendix-include-cobjgetsingle:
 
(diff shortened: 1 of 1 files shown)


sypets:
Releases:

* [x] master
* [ ] 10.4
* [ ] 9.5

PR #295, opened 2020-03-28
Fix references to paragraphs on same page

jonaseberle:
The "same as includeJSLibs" pointed to itself. "Above" was not always pointing to something above but rather below.


DanielSiepmann:
Thank you for fixing this confusing wrong information @jonaseberle.

PR #297, opened 2020-04-02
[TASK] add docker-compose.yml

jonaseberle:
This adds docker-compose.yml as per
https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/RenderingDocs/RenderWithDockerCompose.html

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,8 @@
+version: '2'
+services:
+  t3docmake:
+    image: t3docs/render-documentation:latest
+    volumes:
+    - ./:/PROJECT:ro
+    - ./Documentation-GENERATED-temp:/RESULT
+    command: makehtml
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '2'
+version: '3'
 services:
   t3docmake:
     image: t3docs/render-documentation:latest


docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: '2'
```
mbrodala, line 1: I think we should at least use `'3'` here to have all new features available once we need them.

I've also added a PR for the linked page.

jonaseberledmind: It doesn't change anything (newer docker-compose just interpretes it).
I just took it from here tbh:
https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/RenderingDocs/RenderWithDockerCompose.html#render-with-docker-compose

mbrodala: I know, but why use TYPO3v8 when you can use TYPO3v9? ;-)

Also see https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/pull/151

jonaseberledmind: ok

sypets:
Hi, what is the status on this? Are the requested changes resolved?

btw, I am not up to date on the status (haven't had time to read Slack lately. I would suggest to make a general decision about docker-compose.yml, refine the docker-compose.yml in "Writing Documentation" and then use that as basis. This can be pretty quickly added to the rest of the repositories. 

If it helps, we could also use the decision platform or have a Slack call. 

On the subject matter, I cannot help much. I use the proposed Docker commands (without docker-compose). It works fine on my system, but I am aware that some people with Win and Mac have had problems with the commands and the docker-compose solution works better for them. I would also suggest to contact @marble

jonaseberle:
The requested changes are resolved.

This is just a starting point. While I am quite sure this already helps some people, single commands could be predefined just as in the current bash script.

So I am open to a general discussion.

sypets:
Please take this up for discussion in Slack channel.

marble:
@jonaseberle , what os are you using? Windows, Mac, Linux?

jonaseberle:
@marble I am using Linux and Windows

marble:
For many reasons this should not go in. Nor should it be added to other manuals:

- The file does add content to the manual
- It only adds a generic convenience file which can easily copied there
- We (team) should not give us a hard time by opening up a field for discussions about rendering convenience.
- Personally I think it's the wrong solutions to use Docker compose to bake in _only some_ of the possible arguments into  a command
- If somebody thinks he doesn't want to use the officially supported "dockrun" solution s-he could write create the same functionality of this Docker compose solution with a bash of bat or powershell script of a few lines.
- It took a while to get rid of the `_make` folders in the repos of the early times - let's not start with this again
- For your info: I have already written a `dockrun_t3rd.bat` script for Windows with the same capabilities. However, no matter how hard I tried, up to now I could find a single Windows user who would be willing to try the script and give feedback.
- If somebody doesn't want to use the existing `dockrun` script becaus s-he doesn't trust them - well: ① No other users has reported similar concerns so far, ② s-he may help to improve the script or read through it.

marble (changes requested):
Do not merge. Read my comments.

PR #300, opened 2020-04-15
Corrcted Example, colPos is not in .select

linawolf:
There is no such thing as colPos in .select, see https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/Functions/Select.html for Reference


sypets:
Releases:

* [x] master
* [ ] 10.4
* [ ] 9.5

sypets:
Backported

PR #302, opened 2020-04-27
cObj FILE was removed with TYPO3 10

froemken:


sypets:
Ah, I did not see this PR. There is now another PR with some more changes (other occurrences of FILE).

Since your PR was first, I will merge here, but add the other changes on top first.

sypets:
See also my question about how to use the property ".template" in FLUIDTEMPLATE or TEMPLATE. As FILE no longer exists .template does not work as previously, perhaps it should get removed entirely.

I now added a warning about this.

https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/issues/314

sypets:
Backported to 10.4

PR #303, opened 2020-04-28
Remove out of date information about the scope of the `backend` symfony expression

m41r:
Due to https://review.typo3.org/c/Packages/TYPO3.CMS/+/63752 the statement `available in BE only` is not true anymore


sypets:
Releases:

* [ ] master
* [x] 10.4
* [ ] 9.5

PR #308, opened 2020-05-20
Add hint for 32bit operating systems

peterkraume:
In `typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php` `cObjGet()` filters and sorts the array keys with `filterAndSortByNumericKeys()` and the help of `MathUtility::canBeInterpretedAsInteger()`.

This is done by a type cast to `int` and a comparison with the original key. This type cast fails with numbers greater than `2147483647` on 32bit operation systems. Yes, there are still web hosting companies out there with 32bit operation systems!

### DIFF
--- a/Documentation/Setup/Page/Index.rst
+++ b/Documentation/Setup/Page/Index.rst
@@ -310,7 +310,7 @@ cssInline.[array]
    Description
          Allows to add inline CSS to the page :html:`<head>` section.
          The :ts:`cssInline` property contains any number of numeric keys, each representing one cObject.
-         The numeric key should not exceed 2147483647 to prevent problems on 32bit operating systems.
+         Internally handled as PHP integer, maximum number is therefore restricted to `PHP_INT_MAX`
 
    Example
          ::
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
Thanks. That's not the only place. What about changing your suggestion into something like: Internally handled as PHP Integer, maximum number is therefore restricted to `PHP_INT_MAX`?

As this is platform dependant, not only 32 or 64 bit: https://www.php.net/manual/en/language.types.integer.php

This also applies to `typeNum` property, as far as I know. At least it did back in the years. But adding in for somre properties is already an improvement worth to merge.

DanielSiepmann:
- [ ] 10.4 (3547e17f0a6f07444abc904c377c2b0d4aa0c1df)
- [ ] 9.5 (f01d67c9da315fc4b328d7585e5548ad1f0a74a4)
- [ ] 8.7 (e3d7fe2f7af82d741c37c8c5c3289ca444c06ab8)

Issue #313, opened 2020-06-17
cssInline sample with FILE, but FILE is deprecated and no longer supported


dirk2014:
I guess there is a small bug in the TYPO3 documentation at

- https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/Setup/Page/Index.html?highlight=cssinline#cssinline-array

There is used a TypoScript snippet with the content object FILE. But FILE is deprecated since 9.5 and isn't supported in TYPO3 10.2. See:

- https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/ContentObjects/File/Index.html

sypets:
Thank you for your report. I created a pull request: #315

In the future, you can also make changes yourself, see https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingDocsOfficial/Index.html

Issue #314, opened 2020-06-20
How to use template in FLUIDTEMPLATE since TYPO3 9


sypets:
This is a question:

Since TYPO3 9.5, the `FILE` content object is deprecated, see https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5/Deprecation-85970-FileContentObject.html

Now, documentation should get changed. `FILE` is still used in `FLUIDTEMPLATE` in [template](https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/ContentObjects/Fluidtemplate/Index.html#template) property and [example](https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/ContentObjects/Fluidtemplate/Index.html#example):

```
page = PAGE
page.10 = FLUIDTEMPLATE
page.10 {
   template = FILE
   template.file = EXT:site_default/Resources/Private/Templates/MyTemplate.html
```

For TEMPLATE it is also used:

```
page.10 = TEMPLATE
page.10 {
   template = FILE
   template.file = fileadmin/template.html
}
```

https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/ContentObjects/Template/Index.html#template

Since FILE is deprecated (and removed), this should get changed, but how?

* does it make sense to still have the `template` property at all (and not use the `file` property  instead)?
* how should the above code snippets look?

Related: #313

sypets:
In fact, in TYPO3 10.4.4, using .file will work:

```
page = PAGE
page.1 = FLUIDTEMPLATE
page.1.file = fileadmin/templates/Fluidtemplate.html

```

But using .template in the documented way causes an exception:

```
page = PAGE
page.1 = FLUIDTEMPLATE
page.1.template = FILE
page.1.template.file = fileadmin/templates/Fluidtemplate.html

```

> Could not find template source for FILE

https://wiki.typo3.org/Exception/CMS/1437420865

sypets:
In TYPO3 11 TEMPLATE will be removed. So we don't have to worry about that then.

For FLUIDTEMPLATE, using .template is not really necessary, since you can use .file instead. Should we just remove it?

linawolf:
Merged

PR #315, opened 2020-06-20
[BUGFIX] Remove FILE from snippet

sypets:
FILE has been deprecated since TYPO3 9. Remove FILE from code snippets
and use alternatives.

Also, add warning to TEMPLATE that it will be removed in TYPO3 11

Resolves: #313
Related: #314
Releases: master, 10.4, 9.5


sypets:
Made redundant with #302

PR #319, opened 2020-06-25
Add example to request.getQueryParams with check if query param is de…

sypets:
Add an example to request.getQueryParams with check if query param is defined

Releases: master, 10.4, 9.5

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -617,9 +617,15 @@ request.getQueryParams()
       Value: ``1``
 
 :aspect:`Example`
-   ::
+   Check if query parameter skipSessionUpdate equals 1::
 
       [request.getQueryParams()['skipSessionUpdate'] == 1]
+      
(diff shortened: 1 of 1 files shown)


sypets:
Some more todos:

* there are better examples, see https://forge.typo3.org/issues/89176

sypets:
This is the same example as found under [traverse](https://docs.typo3.org/m/typo3/reference-typoscript/10.4/en-us/Conditions/Index.html#traverse) which is a bit unfortunate, but I think if someone wants to use getQueryParams and is not aware of this possibility and expects TypoScript to handle the check internally, this is a common error they might run into so it is a good idea to have the example shown for getQueryParams(). 

See also issues:

* [TypoScript – Symfony Expressions: request.getQueryParams() should do an "is defined" validation](https://forge.typo3.org/issues/88756)
* [Provide better way to traverse array in TS conditions](https://forge.typo3.org/issues/89176)

PR #330, opened 2020-08-02
New condition syntax is now mandatory for 10 (10.4)

sypets:
- feature flag TypoScript.strictSyntax is no longer available, change
  text for this
- add information that new conditions are mandatory and add links to
  changelogs
- add short overview of old / new conditions
- refer to deprecation log


sypets:
Before merging to "master" (for version 11) I will shorten the text.

PR #357, opened 2020-09-28
Bugfix: name extTarget correctly

amtee:
This one drove me crazy. :) The property is named `target` but it is ought to be `extTarget`


sypets:
> This one drove me crazy. :) 

I can certainly relate to that. ;-)

Indeed, "target" is twice on that page, the first one was "extTarget" (which is also the case for 9.5 and below). 

I merged and "upported" to master.

sypets:
@amtee Welcome and thank you!

PR #372, opened 2020-11-06
Improve description of the config.no_cache

tmotyl:
can be backported to older branches


Documentation/Setup/Config/Index.rst
```diff
@@ -1522,12 +1522,17 @@ no\_cache
...
+         0
 
    Description
-         If this is set to true, the page will not be cached. If set to false,
-         it's ignored. Other parameters may have set it to true of other
-         reasons.
+         If this is set to 1, it disables the `pages` cache, meaning that the rendered result/response
+         will not be saved to cache.
+         If set to 0, it's ignored. Rendered result (e.g. full html of a page) is stored in the `pages` cache.
+         Other parameters may have set it to true of other reasons.
```
sypets, line 1531: > Other parameters may have set it to true of other reasons.

would use "for other reasons" instead.

mbrodala (approved):

PR #377, opened 2020-12-09
Do not list "request" function as working in all contexts

moe2k:
Request does not work in CLI context, so it shouldn't be listed under "available in all contexts". I didn't test whether the other functions listed there work in CLI context but at least we have one improvement.

Also see https://forge.typo3.org/issues/91654

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -490,11 +490,11 @@ typo3.devIpMask
 
 .. _condition-functions-in-all-contexts:
 
-Functions in all contexts
+Functions in frontend and backend context
 -------------------------
 
 Functions take over the logic of the old conditions which do more than a simple comparison check.
-The following functions are available in **any** context:
+The following functions are available in frontend and backend context, but not CLI.
 
 .. _condition-function-request:
 
@@ -747,6 +747,12 @@ request.getPageArguments()
    
    Allows migration from old condition syntax using `[globalVar = GP:singlepartner > 0]`
    to `[request.getPageArguments().get('singlepartner') > 0]`.
+   
+   
+Functions in all contexts
+-------------------------
+
+The following functions are available in **all** contexts: frontend, backend and CLI
 
 .. _condition-function-date:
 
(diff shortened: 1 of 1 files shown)


Documentation/Conditions/Index.rst
```diff
@@ -490,11 +490,11 @@ typo3.devIpMask
 
 .. _condition-functions-in-all-contexts:
 
-Functions in all contexts
+Functions in frontend and backend context
 -------------------------
```
gilbertsoft, line 494:
```suggestion
Functions in frontend and backend context
-----------------------------------------
```

Documentation/Conditions/Index.rst
```diff
@@ -747,6 +747,12 @@ request.getPageArguments()
    
    Allows migration from old condition syntax using `[globalVar = GP:singlepartner > 0]`
    to `[request.getPageArguments().get('singlepartner') > 0]`.
+   
+   
+Functions in all contexts
+-------------------------
+
+The following functions are available in **all** contexts: frontend, backend and CLI
```
gilbertsoft, line 755:
```suggestion
The following functions are available in **any** context:
```

gilbertsoft: There is also the install tool context, so we'd better not make a list here.

Documentation/Conditions/Index.rst
```diff
@@ -490,11 +490,11 @@ typo3.devIpMask
 
 .. _condition-functions-in-all-contexts:
 
-Functions in all contexts
+Functions in frontend and backend context
 -------------------------
 
 Functions take over the logic of the old conditions which do more than a simple comparison check.
-The following functions are available in **any** context:
+The following functions are available in frontend and backend context, but not CLI.
```
gilbertsoft, line 497:
```suggestion
The following functions are available in frontend and backend context, but not CLI:
```

gilbertsoft (changes requested):

Documentation/Conditions/Index.rst
moe2k: didn't know that - makes sense!

linawolf (approved):

gilbertsoft (approved):

PR #386, opened 2020-12-30
Set max_line_length to 80in .editorconfig as should be the standard

linawolf:
(see https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CodingGuidelines.html?highlight=editorconfig)


flicstar (approved):
Looks good

linawolf:
According to @marble this would cause problems with some editors, therfore we dont merge this

PR #396, opened 2021-01-03
How to use template in FLUIDTEMPLATE with property template

### LINKED ISSUE #314: How to use template in FLUIDTEMPLATE since TYPO3 9

linawolf:
resolves #314

preport to master

I accidently commited this to 9.5. If there are change requests they also need to be merged into 9.5

### DIFF
--- a/Documentation/ContentObjects/Fluidtemplate/Index.rst
+++ b/Documentation/ContentObjects/Fluidtemplate/Index.rst
@@ -35,7 +35,7 @@ The functionality can be further augmented and customized with
 Properties
 ==========
 
-.. contents:: 
+.. contents::
    :local:
    :depth: 1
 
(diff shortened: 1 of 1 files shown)


flicstar (approved):
This looks good to me (but what do I know?) :-)

PR #399, opened 2021-01-15
Update start page

marble:


sypets:
I created a decision about what should be on the start page: https://decisions.typo3.org/t/done-proposal-for-documentation-manual-start-page/668/2 a while ago

proposing to remove the "Rendered today". This is already in the footer and I believe it is unnecessary on the start page. I tried to remove all unnecessary information.

Now the "Rendered today" has been added again (after it was removed a while ago).

Can we please decide these things in general and not make changes back and forth?

I had assumed the changes I proposed had been agreed on and they are already merged in "Writing documentation" as well: https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/ExampleTexts.html#start-page

The same things goes for the sitemap-link on the start page.

It is a lot of work to try to make these decisions and unify manuals. If everyone does it differently, we keep changing things back and forth.

marble:
I don't think it's a decision of life and death weather that "Rendered at" is there or not. Nor do I think it is very important if that is the same on every start page. I've added it, because I did a few quick changes today before the meeting and the rendering wasn't working at Bamboo. I didn't want to scroll down all the time. And what would it hurt? So what? The problem with this kind of "rules" on decisions: They cannot be a stable as a constitutional law. We have to be more flexible. Mostly we are using evolving best practises. For example: You asked about the Sitemap link on the start page. And we "decided" something to have that and have a link to the "link targets" in parallel. Yes, I now dropped both links alltogether from there, because: The sitemap link was created at a time when the sitemap was not part of the normal menu. That's gone, so it isn't needed any more. In the meantime we have come up with the Index page(s), and that is sufficient to have an entry to the "link targets", which in turn is sufficient IMHO, because then it's linked. And we don't need that link on the start page at all.

I'm having a problem with a way you are addressing these things. For me the normal way would be: Ping me, ping us in a less formal way like "Hey, what's going on there, didn't we want to do that differently?" Then this could have be quickly resolved. Now you and me are wasting a lot of time by having a written argument.

PR #403, opened 2021-01-18
why "PAGE & PAGE"?

franzholz:


mbrodala (approved):
Back then this was `page & PAGE`, so the conventional name for this object and its actual cObject type. Of course, this doesn't make sense anymore. ;-)

linawolf:
@mbrodala  you would backport this to which Versions?

mbrodala:
@linawolf 10.4 and 9.5. There `PAGE & page` is still used but the same reasoning for dropping the latter should be applied there.

PR #408, opened 2021-01-28
add example with page id

franzholz:
Only one member variable of $TSFE is listed.
Where should the other parts of $TSFE be described?

---------

    /**
     * The page id (int)
     * @var string
     */
    public $id = '';

    /**
     * The type (read-only)
     * @var int|string
     */
    public $type = '';

    /**
    /**
     * Page will not be cached. Write only TRUE. Never clear value (some other
     * code might have reasons to set it TRUE).
     * @var bool
     */
    public $no_cache = false;

    /**
     * The rootLine (all the way to tree root, not only the current site!)
     * @var array
     */
    public $rootLine = [];

    /**
     * The pagerecord
     * @var array
     */
    public $page = [];

    /**
     * This will normally point to the same value as id, but can be changed to
     * point to another page from which content will then be displayed instead.
     * @var int
     */
    public $contentPid = 0;

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1141,6 +1141,10 @@ getTSFE
 
       [getTSFE().type == 98]
 
+   Current page id equals to 17. :ref:`PAGE`::
+
+      [getTSFE().id == 17]
+
 
 .. index:: Conditions; getenv
 .. _condition-function-getenv:
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1141,7 +1141,7 @@ getTSFE
 
       [getTSFE().type == 98]
 
-   Current page id equals to 17. :ref:`PAGE`::
+   Current page id equals to 17. However the recommended way is with :ref:`_condition-page`::
 
       [getTSFE().id == 17]
 
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1141,7 +1141,7 @@ getTSFE
 
       [getTSFE().type == 98]
 
-   Current page id equals to 17. However the recommended way is with :ref:`condition-page`::
+   Current page id equals to 17. However, :ref:`condition-page` should be preferred::
 
       [getTSFE().id == 17]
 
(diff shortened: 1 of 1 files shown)


Documentation/Conditions/Index.rst
```diff
@@ -1141,6 +1141,10 @@ getTSFE
 
       [getTSFE().type == 98]
 
+   Current page id equals to 17. :ref:`PAGE`::
```
mbrodala, line 1144: Linking to the explanation of the `PAGE` TypoScript object does not add any value here IMO.

mbrodala (changes requested):
Actually `page["uid"]` should be just fine, there is technically no difference between these two. So I think documenting other ways to access the same value will only lead to confusion.

Documentation/Conditions/Index.rst
franzholz: https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/Setup/Page/Index.html

Only here you have an explanation for the id of the page.

franzholz: Before TYPO3 10 you could write the TypoScript Condition like this:

    [globalVar = TSFE:id = 279]

So I find this solution:

    [getTSFE().id == 279]

I could not find your solution at all.

Another hint could be added in order to refer to page based conditions.

mbrodala: Exactly: https://docs.typo3.org/m/typo3/reference-typoscript/10.4/en-us/Conditions/Index.html#page

franzholz: I have added the local link to page instead.

Documentation/Conditions/Index.rst
```diff
@@ -1141,6 +1141,10 @@ getTSFE
 
       [getTSFE().type == 98]
 
+   Current page id equals to 17. However the recommended way is with :ref:`condition-page`::
```
mbrodala, line 1144:
```suggestion
   Current page id equals to 17. However, :ref:`condition-page` should be preferred::
```

mbrodala (changes requested):

mbrodala (approved):

PR #416, opened 2021-02-24
Mention traverse() result on missing array entry

mbrodala:
Releases: master, 10.4, 9.5

Source: https://github.com/TYPO3/TYPO3.CMS/blob/10.4/typo3/sysext/core/Classes/ExpressionLanguage/FunctionsProvider/DefaultFunctionsProvider.php#L128

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1005,13 +1005,15 @@ traverse
    Custom
 
 :aspect:`Description`
-   This function has two parameters:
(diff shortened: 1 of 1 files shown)


susannemoog (approved):

susannemoog:
The whole traverse function is not documented in 9.5 so I did not do that backport.

mbrodala:
Funny, even though it is used here: https://docs.typo3.org/m/typo3/reference-typoscript/9.5/en-us/Conditions/Index.html#request-getqueryparams

But that's OK, thanks for checking!

PR #417, opened 2021-02-25
[TASK] Add some hints for deprecated language settings (9.5)

sypets:
Some language settings have been deprecated in TYPO3 but no deprecated notes added.

### DIFF
--- a/Documentation/Setup/Config/Index.rst
+++ b/Documentation/Setup/Config/Index.rst
@@ -1197,6 +1197,9 @@ htmlTag\_dir
          :ref:`data-type-string`
 
    Description
+         .. deprecated:: 9
+            See note below.
+
          Sets text direction for whole document (useful for display of Arabic,
          Hebrew pages).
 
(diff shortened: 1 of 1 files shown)


Documentation/Setup/Config/Index.rst
```diff
@@ -1625,6 +1636,10 @@ locale\_all
          :ref:`data-type-string`
 
    Description
+        This setting will be removed in v10. In v9, language settings
+        from the :ref:`Site configuration <t3coreapi:sitehandling>` are used by default and the TypoScript settings
```
sypets, line 1640: It might make more sense to link to this page: https://docs.typo3.org/m/typo3/reference-coreapi/9.5/en-us/ApiOverview/SiteHandling/AddLanguages.html, rather than to the general Site handling page.

linawolf: I think we should use the `deprecated` directive here

sypets:
I accidentally deleted a comment. Meant to delete my own comment only. Is resolved now (I added the missing "deprecated" directives.

marble (approved):
Good work! Trusting you that content is correct.

PR #420, opened 2021-02-28
add .after and .afterWrap properties

erredeco:
Add description to .after and .afterWrap properties (that seemed absent in documentation since 4.5?)  in order to solve https://forge.typo3.org/issues/91965

Please review it! I am not sure if I put them in the correct place that respects the TypoScript execution order!

Please backport it if you want

### DIFF
--- a/Documentation/MenuObjects/Tmenuitem/Index.rst
+++ b/Documentation/MenuObjects/Tmenuitem/Index.rst
@@ -76,8 +76,27 @@ stdWrap.data = field : [field name].
 
    Description
          wrap around the ".before"-code
+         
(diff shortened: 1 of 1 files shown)


linawolf (approved):

robinson2:
This looks good. It would be nice if "after" still had a description. Maybe we can just take the one from /sysext/t3editor/Resources/Private/tsref.xml:

The series of "before..." properties are duplicated to "after..." properties as well. The only difference is that the output generated by the .after.... properties are placed after the link and not before

Unfortunately, "before" does not have a description yet and there is none in the file mentioned above that we could copy.

PR #424, opened 2021-03-23
Add recursive example for gettext DB

linawolf:
Releases: master, 10.4, 9.5

### DIFF
--- a/Documentation/DataTypes/Properties/GetText.rst
+++ b/Documentation/DataTypes/Properties/GetText.rst
@@ -147,7 +147,12 @@ DB
    Get the value of the header field of record with uid 234 from table
    "tt\_content"::
 
-      foo = DB : tt_content:234:header
+      foo.data = DB : tt_content:234:header
+      
+   Get the value of the header field of a record, whose uid is stored in a GET
+   parameter `myContentId`::
+   
+      foo.data.dataWrap = DB : tt_content:{GP : myContentId}:header
 
 .. _data-type-gettext-debug:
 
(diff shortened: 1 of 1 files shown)


Documentation/DataTypes/Properties/GetText.rst
```diff
@@ -147,7 +147,12 @@ DB
    Get the value of the header field of record with uid 234 from table
    "tt\_content"::
 
-      foo = DB : tt_content:234:header
+      foo.data = DB : tt_content:234:header
+      
+   Get the value of the header field of a record, whose uid is stored in a GET
+   parameter `myContentId`::
+   
+      foo.data.dataWrap = DB : tt_content:{GP : myContentId}:header
```
susannemoog, line 155: are we sure that this is not an SQL injection? From the top of my head I'm unsure if that's validated to int internally or not.

mbrodala, line 155: We should mention that it is safe to directly use a client-/user-provided value here since internally the UID will be cast to integer and used as named parameter in the query.

This should raise awareness of SQL injections and hopefully avoid reports about a potential security issue here.

mbrodala: Hehe, see my comment below. ;-)

susannemoog: ha, perfect :) thanks.

linawolf: Thanks, I added the information, this is important

susannemoog (approved):

PR #433, opened 2021-04-21
Update of page array example

lukasdoerr:
For better understanding how to make conditions with it

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -78,7 +78,7 @@ page
       [page["backend_layout"] == 5]
       [page["backend_layout"] == "example_layout"]
       
-      #Maybe you want to check the page title 
+      # Check the page title 
       [page["title"] == "foo"]
 
 .. index:: Conditions; Constant
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -67,7 +67,7 @@ page
    Array
 
 :aspect:`Description`
-   You can read all page information from the page array. To find out, what variable has which name, you can simply check the input fields in backend by inspecting the input name. Or you can enable BE Debug Mode, to see the name of the variables.
+   All data of the current page record as array. To find out which fields are available, you can enable the debug mode in the TYPO3 backend which will display the field names.
 
 :aspect:`Example`
    ::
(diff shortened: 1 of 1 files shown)


tomwarwick (approved):

Documentation/Conditions/Index.rst
```diff
@@ -67,13 +67,21 @@ page
...
       [page["uid"] == 2]
-
+      
+      # Or when you need the page backend layout in V9.4 or V10:
+      [page["backend_layout"] == 5]
+      
+      # Or the backend layout has a name:
+      [page["backend_layout"] == "example_layout"]
+      
+      #Maybe you want to check the page title
```
mbrodala, line 83:
```suggestion
      # Check the page title 
```

Documentation/Conditions/Index.rst
```diff
@@ -67,13 +67,21 @@ page
...
-
+      
+      # Or when you need the page backend layout in V9.4 or V10:
+      [page["backend_layout"] == 5]
+      
+      # Or the backend layout has a name:
+      [page["backend_layout"] == "example_layout"]
+      
+      #Maybe you want to check the page title 
+      [page["title"] == "foo"]
```
mbrodala, line 84: TBH this example could be improved. The title is highly dependent on user input and can be changed any time. So this is not something code can rely on.

mbrodala (changes requested):

Documentation/Conditions/Index.rst
lukasdoerr: Problem with these docs is, it is highly intransparent, where to find information, nor there is some information for it to find out, where to find the right use for it. But sure, the text can be minidied like this. im totally okay with that. I really just want to show (As a person who is new to TYPO3) how to easily access page variables.

mbrodala: Technically you need to look at the schema of the `pages` table in your database since that's the definitive source. Even the backend form or TCA won't give you the full lists of fields.

mbrodala (approved):

Documentation/Conditions/Index.rst
lukasdoerr: And thats what i mean, that info for the schema of "pages" should be linked in the docs as well. Is there a full list for that? I can update that as well. I think that would really help, finding out, what potential "page" array has.

mbrodala: No, that's not possible since it depends on the extensions you have installed and activated. Any extension can adjust the schema of any table to add or even change fields.

PR #438, opened 2021-05-10
[BUGFIX] Use traverse() with getPageArguments()

mabolek:
`getPageArguments()` returns an object implementing the `ArrayAccess` interface, and it doesn't have a `get()` method. The `ArrayAccess` method `offsetGet` could probably be used, but it doesn't have the same error-silencing behavior as `traverse()`.

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -970,10 +970,11 @@ request.getPageArguments()
    Get current `PageArguments` object with resolved route parts from enhancers.
 
 :aspect:`Example`
-   [request.getPageArguments().get('foo_id') > 0]
+   [traverse(request.getPageArguments(), 'foo_id') > 0]
 
    Allows migration from old condition syntax using `[globalVar = GP:singlepartner > 0]`
-   to `[request.getPageArguments().get('singlepartner') > 0]`.
+   to `[traverse(request.getPageArguments(), 'singlepartner') > 0]` (See 
+   :ref:`condition-function-traverse`).
 
 
 .. index:: Conditions; date


Documentation/Conditions/Index.rst
```diff
@@ -970,10 +970,11 @@ request.getPageArguments()
    Get current `PageArguments` object with resolved route parts from enhancers.
 
 :aspect:`Example`
-   [request.getPageArguments().get('foo_id') > 0]
+   [traverse(request.getPageArguments(), 'foo_id') > 0]
```
mbrodala, line 973: Unfortunately `traverse()` currently only supports `array`, no `ArrayObject` or similar. While this by itself could be improved actually we have a slight inconsistency here.

The `RequestWrapper` for expressions hints at the `PageArguments` class (which implements `RouteResultInterface`) but technically it could also return a `SiteRouteResult` (same).

So this should be cleared up first IMO.

mbrodala: But `PageArguments::get()` exists:

https://github.com/TYPO3/TYPO3.CMS/blob/10.4/typo3/sysext/core/Classes/Routing/PageArguments.php#L123

So technically it is mostly correct aside from the inconsistency I mentioned.

mabolek: @mbrodala Do you know where I can find the implementation of the `traverse()` method?

mbrodala: Sure:

https://github.com/TYPO3/TYPO3.CMS/blob/7d9aa7580de50da491308a5786a612ddba376da0/typo3/sysext/core/Classes/ExpressionLanguage/FunctionsProvider/DefaultFunctionsProvider.php#L117

linawolf:
Has been on draft since a year. Closing.

PR #441, opened 2021-06-09
fix addQueryString option in HMENU language

akiessling:
addQueryString.method needs to be set to GET to properly work in the language menu, at least in v10
Without it, no params are added to the link

### DIFF
--- a/Documentation/ContentObjects/Hmenu/Index.rst
+++ b/Documentation/ContentObjects/Hmenu/Index.rst
@@ -1518,6 +1518,7 @@ Creates a language menu:
          parameter.data = page:uid
          additionalParams = &L=0 || &L=1 || &L=2
          addQueryString = 1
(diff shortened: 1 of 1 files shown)


linawolf:
addQueryString.method defaults to the empty string, which results in using GeneralUtility::_GET(), See https://review.typo3.org/c/Packages/TYPO3.CMS/+/61295/12/typo3/sysext/extbase/Classes/Mvc/Web/Routing/UriBuilder.php#320

If this is not working please make a bug report at https://forge.typo3.org/projects/typo3cms-core/issues

linawolf:
I put your PR on DRAFT. If it should turn out that this is a feature, not a bug indeed we can still reopen and merge it. Otherwise please close

linawolf:
Closed as there were no more changes

PR #444, opened 2021-06-16
Add note for pidInList validation

SZChimp:
When I worked with this property, I always got the line 
```sql
(`<table_name>`.`uid`= 0)
``` 
inserted into my query and it took me a long time to figure out, where this statement was from.
When you specify a bad doktype page id, which does not meet the criteria for [this method](https://github.com/TYPO3/TYPO3.CMS/blob/10.4/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L6765), the line is inserted.

Please add more notes like these.

### DIFF
--- a/Documentation/Functions/Select.rst
+++ b/Documentation/Functions/Select.rst
@@ -154,7 +154,18 @@ pidInList
       Allows to disable the :sql:`pid` constraint completely. Requirements: 
       :ts:`uidInList` *must* be set or the table *must* have the prefix 
       "static\_\*".
+      
+   .. note::
+      Check the doktype of your backend page. If you are trying to fetch records from
+      e.g. a folder, the :php:`$cObj->checkPid_badDoktypeList` method will insert the
+      following SQL into your query:
 
+      .. code-block:: sql
+
+      [...]WHERE (`your_requested_table_name`.`uid` = 0) AND [...]
+
+      Which will always result in an empty query result!
+   
 :aspect:`Default`
    :ts:`this`
 
--- a/Documentation/Functions/Select.rst
+++ b/Documentation/Functions/Select.rst
@@ -157,14 +157,14 @@ pidInList
       
    .. note::
       Check the doktype of your backend page. If you are trying to fetch records from
-      e.g. a folder, the :php:`$cObj->checkPid_badDoktypeList` method will insert the
+      e.g. a sys_folder, the :php:`$cObj->checkPid_badDoktypeList` method will insert the
       following SQL into your query:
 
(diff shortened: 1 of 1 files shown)


Documentation/Functions/Select.rst
```diff
@@ -154,7 +154,18 @@ pidInList
...
+   .. note::
+      Check the doktype of your backend page. If you are trying to fetch records from
+      e.g. a folder, the :php:`$cObj->checkPid_badDoktypeList` method will insert the
+      following SQL into your query:
 
+      .. code-block:: sql
+
+      [...]WHERE (`your_requested_table_name`.`uid` = 0) AND [...]
+
+      Which will always result in an empty query result!
```
linawolf, line 167: Please try formulating this without exclamation mark...

Documentation/Functions/Select.rst
```diff
@@ -154,7 +154,18 @@ pidInList
       Allows to disable the :sql:`pid` constraint completely. Requirements: 
       :ts:`uidInList` *must* be set or the table *must* have the prefix 
       "static\_\*".
+      
+   .. note::
+      Check the doktype of your backend page. If you are trying to fetch records from
+      e.g. a folder, the :php:`$cObj->checkPid_badDoktypeList` method will insert the
```
linawolf, line 160: by folder you mean a page with type sys_folder?

Documentation/Functions/Select.rst
```diff
@@ -154,7 +154,18 @@ pidInList
       Allows to disable the :sql:`pid` constraint completely. Requirements: 
       :ts:`uidInList` *must* be set or the table *must* have the prefix 
       "static\_\*".
+      
+   .. note::
```
linawolf, line 158: did you find a better solution? Should we point people there instead?

linawolf (approved):

Documentation/Functions/Select.rst
SZChimp: No, I just think it should be stressed to check the page to be of the correct type.

Documentation/Functions/Select.rst
SZChimp: I rephrased it, you're right. It looks odd :)

Documentation/Functions/Select.rst
SZChimp: Yes, I updated the word to be more specific, thank you.

linawolf (approved):

linawolf:
Releases: master, 10.4

PR #453, opened 2021-08-04
Update Examples.rst

griebie:
I came across this, because I thought at first the code snippet creates the JSON view, which isn't the case. If the JSON view comes along with TYPO3, it must be written this way, I am pretty sure :).
Regards, Thorsten


linawolf (approved):

linawolf:
Thanks for catching this

Greetings Lina

Issue #457, opened 2021-09-08
roundtrip includeJSFooterlibs, includeJSlibs


DavidBruchmann:
The two detail blocks
 - https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/Setup/Page/Index.html#setup-page-includejsfooterlibs-array  
 - https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/Setup/Page/Index.html#setup-page-includejslibs-array

link to each other but the details are nowhere explained.  
So the parameters `external`, `integrity`, `crossorigin` and perhaps more are just missing.
Instead the links should point to https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/Setup/Page/Index.html#includejs-array where everything is explained.

PR #546, opened 2022-06-16
[TASK] Add properties to includeJSlibs and includeJSFooterlibs

brotkrueml:
Releases: main, 11.5, 10.4
Resolves: #457


linawolf (changes requested):
Please let us not include new rst.txt includes that are included serveral times and not auto-generated. They have several drawbacks

- duplicate content possibly
- cannot be edited with the edit on Github workflow
- You would have to manually compare if they are really the same options on reading

Let us rather put the properties with the more common of the properties and link from the other one.

linawolf (approved):

PR #553, opened 2022-07-01
uidInList requires pidInList to be set

linawolf:


Documentation/Functions/Select.rst
```diff
@@ -117,12 +117,20 @@ uidInList
 
    **Note:** :typoscript:`this` is a *special keyword* and replaced with the id of the
    *current record*.
+   
+   .. important:: 
+      :ref:`pidInList` has to be set or there will be no results. Setting
```
mbrodala, line 122: Maybe we should emphasize the default `this` for `pidInList` instead. There is actually no relation between these two options, both have their own unrelated behavior.

linawolf: Ok, well I never had this case before. Did you verify? Usually records are stored in sys_folders while the DataProcessor is run on pages. So I see almost no case where setting uidInList without pidInList with results

linawolf: Ok, so I added the default value to pidInList (as we didn't have it documented yet) and changed the important text accordingly

Documentation/Functions/Select.rst
```diff
@@ -117,12 +117,21 @@ uidInList
 
    **Note:** :typoscript:`this` is a *special keyword* and replaced with the id of the
    *current record*.
+   
+   .. important:: 
+      :ref:`pidInList` is set to :typoscript:`this`. Therefore by defalault only records
+      from the current page are available for uidInList. Setting
+      :typoscript:`pidInList = 0` explicitely disables the check for the pid.
```
mbrodala, line 124:
```suggestion
      :ref:`pidInList` defaults :typoscript:`this`. Therefore by default only records
      from the current page are available for :typoscript:`uidInList`. If records
      should be fetched globally, :typoscript:`pidInList = 0` should also be set.
```

mbrodala (changes requested):

Documentation/Functions/Select.rst
```diff
@@ -117,12 +117,21 @@ uidInList
 
    **Note:** :typoscript:`this` is a *special keyword* and replaced with the id of the
    *current record*.
+   
+   .. important::
```
mbrodala, line 121:
```suggestion
   .. attention:: 
```

As we know, we should avoid `important` . ;-)

mbrodala (approved):

PR #565, opened 2022-07-16
Unify layout of all ContentObjects Part 1

linawolf:
- Use Headlines
- Content Menü
- Sections for Properties and Examples

### DIFF
--- a/Documentation/ContentObjects/Case/Index.rst
+++ b/Documentation/ContentObjects/Case/Index.rst
@@ -26,8 +26,56 @@ be used as it has a special meaning: If the value of the "key"
 property is *not* found in the array of cObjects, then the cObject
 from the "default" property will be used.
 
+.. contents::
+   :local:
 
 .. ### BEGIN~OF~TABLE ###
+
+Properties
+==========
+
+
+1,2,3,4...
+------------
+
+.. container:: table-row
+
+   Property
+         *(array of cObjects)*
+
+   Data type
+         :ref:`cObject <data-type-cobject>`
+
+   Description
+         Array of cObjects. Use this to define cObjects for the different
+         values of "key". If "key" has a certain value, the according
+         cObject will be rendered. The cObjects can have any name, but not
+         the names of the other properties of the cObject CASE.
+
+cache
+-----
+
+.. include:: /DataTypes/Properties/Cache.rst.txt
+
+default
+-------
+
+.. container:: table-row
+
+   Property
+         default
+
+   Data type
+         :ref:`cObject <data-type-cobject>`
+
+   Description
+         Use this to define the rendering for *those* values of "key" that
+         do *not* match any of the values of the "array of cObjects". If no
+         default cObject is defined, an empty string will be returned for
+         the default case.
+
+if
+--
 
 .. container:: table-row
 
@@ -40,18 +88,8 @@ from the "default" property will be used.
    Description
          If "if" returns false, nothing is returned.
 
-
-.. container:: table-row
-
-   Property
-         setCurrent
-
-   Data type
-         string /:ref:`stdWrap <stdwrap>`
-
-   Description
-         Sets the "current"-value.
-
+key
+---
 
 .. container:: table-row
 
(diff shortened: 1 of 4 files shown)


Documentation/ContentObjects/Case/Index.rst
```diff
@@ -26,32 +26,70 @@ be used as it has a special meaning: If the value of the "key"
...
+.. contents::
+   :local:
 
 .. ### BEGIN~OF~TABLE ###
 
+Properties
+==========
+
+
+1,2,3,4...
```
brotkrueml, line 38: As I understand, that can also be strings as it is the value of the selected key. I would avoid 1,2,3,4... as it is near the default usage of a COA, for example (first I thought about that). I would leave it to the previous "(array of cObjects)"

Documentation/ContentObjects/CoaAndCoaInt/Index.rst
```diff
@@ -107,7 +130,8 @@ Examples:
      30.value = <footer>Footer text</footer>
    }
 
-The previous example will print a simple h1 header, followed by the page content records and a footer element.
+The previous example will print a simple h1 header, followed by the page
```
brotkrueml, line 133:
```suggestion
The previous example will print a simple :html:`<h1>` header, followed by the page
```

Documentation/ContentObjects/CoaAndCoaInt/Index.rst
```diff
@@ -107,7 +130,8 @@ Examples:
...
+content records and a footer element.
```
brotkrueml, line 134:
```suggestion
content records and a :html:`footer` element.
```

Documentation/ContentObjects/Content/Index.rst
```diff
@@ -311,19 +202,130 @@ stdWrap
...
+
+CONTENT explained in detail
+----------------------------
+
+See PHP class :php:`\TYPO3\CMS\Frontend\ContentObject\ContentContentObject`
+for details on code level.
+
+Preamble:
+
+.. code-block:: typoscript
```
brotkrueml, line 224: This looks strange as a code block, maybe an admonition is a better way to emphasize the importance of that note?

Documentation/ContentObjects/Content/Index.rst
```diff
@@ -335,6 +337,7 @@ Here is an example of the CONTENT object:
    1.select {
       pidInList = this
       orderBy = sorting
+      where = {#colPos}=0
    }
 
 Since in the above example .renderObj is not set explicitly, TYPO3
```
brotkrueml, line 343:
```suggestion
Since in the above example `.renderObj` is not set explicitly, TYPO3
```

Documentation/ContentObjects/Files/Index.rst
```diff
@@ -8,8 +8,17 @@
 FILES
 =====
 
-A content object of type FILES was integrated with the File Abstraction Layer (FAL)
-and is there to output information about files.
+A content object of type FILES was integrated with the File Abstraction Layer
```
brotkrueml, line 11:
```suggestion
A content object of type FILES uses the File Abstraction Layer
```

Documentation/ContentObjects/Files/Index.rst
```diff
@@ -8,8 +8,17 @@
...
+(FAL) and is there to output information about files.
```
brotkrueml, line 12:
```suggestion
(FAL) and is used to display information about files.
```

Documentation/ContentObjects/Files/Index.rst
```diff
@@ -273,13 +304,13 @@ Special key: "references"
 .. _cobj-files-examples:
 
 Examples:
```
brotkrueml, line 306:
```suggestion
Examples
```

Documentation/ContentObjects/Files/Index.rst
```diff
@@ -273,13 +304,13 @@ Special key: "references"
...
-"""""""""
+==========
```
brotkrueml, line 307:
```suggestion
=========
```

Documentation/ContentObjects/Case/Index.rst
linawolf: true, I mixed this up with the COA

Documentation/ContentObjects/Content/Index.rst
linawolf: I think we don't need that anymore. CONTENT is not really the place to explain what TypoScript is in general. This is probably very old and has been copied ever since

linawolf: removed it

brotkrueml (approved):

PR #570, opened 2022-08-06
Update Index.rst

blueamerican:
remove wrong Anführungszeichen

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -152,7 +152,7 @@ Constant
 
    If constant is a string put constant in quotes::
 
-      ["{$foo.bar}" == "4711"]
+      [{$foo.bar} == "4711"]
 
 .. _condition-tree:
 


tomwarwick:
Hi Markus, have you tested this? I believe the quotation marks are required.

blueamerican:
@tomwarwick 

Yes, i testet it. The qotation marks are right on the right side, not on the left.

brotkrueml:
This change does not fit to the paragraph above: "If constant is a string put constant in quotes:" Both cases are handled for now in the docs: Is value an integer, use no quotes, is value a string, use quotes for both sides.
Will investigate. If applicable, has also be changed in 11.5 and main.

blueamerican:
Quotations on the left side do not work in V10 nor in V11. I tested it.

brotkrueml:
Okay, I testet the following in v10 and v11:

Constants in root template:
```
foo.bar = 4711
```
Conditions in setup in root template:
```
[{$foo.bar} == 4711]
["{$foo.bar}" == 4711]
[{$foo.bar} == "4711"]
["{$foo.bar}" == "4711"]
```

Every condition works correctly. @blueamerican : **So, can you give us your use case, where this does not work?**

But I agree: The differentiation between integer and string is not necessary.

blueamerican:
ah, ist was my mistake.

My constant was
foo = "bar"

As a result, this does not work:

["{$foo}" == "bar"]

Interestingly, then this works:

[{$foo} == "bar"]

But anyway, your're right.. sorry for the confusion.

PR #571, opened 2022-08-08
[TASK] Remove outdated menuitems

linawolf:
move generalitemstates and general menu properties to TMENU, as that is the only surviving menu


brotkrueml:
It is intended hat TMENU.sectionIndex is a separate page and not included into the TMENU page?

Documentation/MenuObjects/Tmenu/Index.rst
```diff
@@ -9,11 +9,235 @@ TMENU
...
+
+   Description
+         (See above)
+
+
+.. ###### END~OF~TABLE ######
+
+[tsref:(cObject).HMENU.(mObj\_itemStates)]
+
+Order of priority: USERDEF2, USERDEF1, SPC, USR, CURIFSUB, CUR,
```
brotkrueml, line 224: I would like to see this as a note, so it is more prominent (either here or at the beginning of the item states list).

linawolf:
> It is intended hat TMENU.sectionIndex is a separate page and not included into the TMENU page?

Right, that doesn't really make sense anymore, I will move it

Documentation/MenuObjects/Tmenu/Index.rst
```diff
@@ -125,24 +360,342 @@ TMENU
...
+.. container:: table-row
+
+   Property
+         sectionIndex
+
+   Data type
+         *(array of keys)*
+
+   Description
+         (see below)
```
brotkrueml, line 395: "see below" is not correct as it is by now a separate page. So either include it into this page or link it accordingly to the other page.

brotkrueml:
Apart from the comments, it is okay for me. In a second PR we can format some stuff accordingly (like HTML as HTML and so on). Can take care of this after merging of this PR.

Documentation/MenuObjects/Tmenu/Index.rst
```diff
@@ -9,11 +9,235 @@ TMENU
 
 .. note::
 
-   TMENU is a menu object type.
+   :typoscript:`TMENU` is a menu object type. In TYPO3 10 the remaining other menu types
```
brotkrueml, line 12:
```suggestion
   :typoscript:`TMENU` is a menu object type. In TYPO3 v10 the remaining other menu types
```

brotkrueml (approved):

PR #582, opened 2022-08-24
Update Index.rst

pi-phi:
replace invalid option 'noLink' with 'doNotLinkIt'


brotkrueml (approved):

brotkrueml:
Ah, it is in branch v9.5. Will take care to add it to the recent branches.

PR #583, opened 2022-08-26
[TASK] Add stdWrap example to FLUIDTEMPLATE paths

brotkrueml:
Additionally, adjust the extension names to reflect the caption of
the snippet and use the Overrides postfix for paths overriding
a template as best practise.

Related: https://forge.typo3.org/issues/98200

Releases: main, 11.5


linawolf:
See the first comment from Oli, it seems like it is not the recommended way so I would not prominently put it in the example. I would rather have it as side-note and also mention the recommended way using conditions

PR #599, opened 2022-09-19
Add reference to changelog of `tree.pagelayout` condition

eliashaeussler:
With this PR, a reference to the changelog entry of the `tree.pagelayout` condition is added.

Releases: main, 11.5

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -187,6 +187,7 @@ tree.pagelayout
 """""""""""""""
 
 .. versionadded:: 11.0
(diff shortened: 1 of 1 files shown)


brotkrueml:
@eliashaeussler Mostly we don't add links to the changelog as the change is described already in the docs (like here). What is the intention and benefit?

eliashaeussler:
Hi @brotkrueml, that's totally fine for me. I was just looking for the condition and thought that it might be useful to have the changelog referenced here to get further information about it. Will close the PR then :)

PR #633, opened 2022-11-09
Add example to remove the website title

Tim-Obert:
There is no config setting for removing the website title from the title tag.  With this workaround you can remove the website title and keep the page title, that also can be set by PageTitleProviders.

### DIFF
--- a/Documentation/Setup/Config/Index.rst
+++ b/Documentation/Setup/Config/Index.rst
@@ -1793,7 +1793,7 @@ pageTitleSeparator
             config.pageTitleSeparator = *
             config.pageTitleSeparator.noTrimWrap = |||
             
-         If you want to get rid of the web page title, you must choose a separator that is not included in the web page title. 
+         If you want to remove the web page title from the displayed title, choose a separator that is not included in the web page title. 
          You then split the title from that character and return everything after it:
 
          .. code-block:: typoscript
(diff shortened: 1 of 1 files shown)


Documentation/Setup/Config/Index.rst
```diff
@@ -1792,6 +1792,20 @@ pageTitleSeparator
 
             config.pageTitleSeparator = *
             config.pageTitleSeparator.noTrimWrap = |||
+            
+         If you want to get rid of the web page title, you must choose a separator that is not included in the web page title.
```
linawolf, line 1796:
```suggestion
         If you want to remove web page title from the displayed title, choose a separator that is not included in the web page title. 
```

Documentation/Setup/Config/Index.rst
```diff
@@ -1792,6 +1792,20 @@ pageTitleSeparator
 
             config.pageTitleSeparator = *
             config.pageTitleSeparator.noTrimWrap = |||
+            
+         If you want to get rid of the web page title, you must choose a separator that is not included in the web page title. 
+         You then split the title from that character and return everything after it:
```
linawolf, line 1797:
```suggestion
         Then split the title from that character and return the second part only:
```

linawolf (approved):

PR #637, opened 2022-11-11
[BUGFIX] Fix code example indentation

eliashaeussler:
Releases: main, 11.5

### DIFF
--- a/Documentation/Setup/Page/Index.rst
+++ b/Documentation/Setup/Page/Index.rst
@@ -239,8 +239,7 @@ bodyTag
    Description
          Body tag on the page
 
-   Example :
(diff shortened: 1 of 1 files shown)


linawolf:
I cannot see any change to the indentation, you removed the double colon?

eliashaeussler:
> I cannot see any change to the indentation, you removed the double colon?

You're right, it's not the indentation which causes the wrong rendering, but rather the colon (or even just the additional blank line?). Currently it looks like this:

image

I streamlined the example with the other examples, which should lead to a proper rendering.

brotkrueml (approved):

PR #646, opened 2022-11-26
[TASK] Add example with path to stdWrap.insertData

brotkrueml:
This is helpful when migrating to typo3/cms-composer-installers v4+.

Related: https://forge.typo3.org/issues/99203
Releases: main, 11.5

### DIFF
--- a/Documentation/Functions/Stdwrap.rst
+++ b/Documentation/Functions/Stdwrap.rst
@@ -1767,6 +1767,11 @@ insertData
       10.value = This is the page title: {page:title}
       10.stdWrap.insertData = 1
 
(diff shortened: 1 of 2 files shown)


Documentation/Functions/Stdwrap.rst
```diff
@@ -1767,6 +1767,11 @@ insertData
       10.value = This is the page title: {page:title}
       10.stdWrap.insertData = 1
 
+      # TEXT is already stdWrap'able, so we can also use insertData right away
```
linawolf, line 1770:
```suggestion
      # TEXT is already stdWrapable, so we can also use insertData right away
```

linawolf (approved):

PR #649, opened 2022-11-27
[FEATURE] #99194 - Extend stdWrap.if feature

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/274

### DIFF
--- a/Documentation/Functions/If.rst
+++ b/Documentation/Functions/If.rst
@@ -215,6 +215,108 @@ equals
       if.value.data = GETENV:REQUEST_METHOD
 
 
+..  index:: if; contains
+..  _if-contains:
+
+contains
+--------
+
+..  versionadded:: 12.1
+
+:aspect:`Property`
+    contains
+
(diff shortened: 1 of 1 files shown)


Documentation/Functions/If.rst
```diff
@@ -215,6 +215,108 @@ equals
...
+                if.contains = Media
+            }
+            outerWrap = <h1>|</h1>
+        }
+
+
+..  index:: if; startsWith
+..  _if-startsWith:
+
+startsWith
```
linawolf, line 259: Should we order them alphabetically or is this a logical sorting regarding order of execution or some such?

brotkrueml: Hmm, I followed the order in the code like the other properties. But can sort it alphabetically in a follow-up. Order does not matter for "if" (like in stdWrap AFAIK) as all conditions have to evaluate to true.

linawolf (approved):

PR #651, opened 2022-11-27
[TASK] Remove "simply" and "just" where not appropriate

brotkrueml:
Inspired by: https://twitter.com/marcelpociot/status/1594706393696985088

Also, some of the usages are just fillers which does not convey meaning.

Releases: main, 11.5

### DIFF
--- a/Documentation/Functions/Calc.rst
+++ b/Documentation/Functions/Calc.rst
@@ -19,7 +19,7 @@ Calculating values (+calc)
 Sometimes a data type is set to ``someType +calc``. The ``+calc`` indicates
 that the value is calculated with ``+-/\*`` operators. *Be aware that the
 operators have no "weight".* The calculation is done from left to
-right.
+right instead of order of operations (multiplication and division before addition and subtraction).
 
 
 How value is calculated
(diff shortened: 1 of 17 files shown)


Documentation/Functions/Calc.rst
```diff
@@ -18,7 +18,7 @@ Calculating values (+calc)
 
 Sometimes a data type is set to ``someType +calc``. The ``+calc`` indicates
 that the value is calculated with ``+-/\*`` operators. *Be aware that the
-operators have no "weight".* The calculation is just done from left to
+operators have no "weight".* The calculation is done from left to
 right.
```
linawolf, line 22:
```suggestion
right instead.
```
(the just was in comparision to the operator weight, this meaning should be kept)

brotkrueml: Why "instead"? It is done from left to right.

linawolf: Yes from left to right instead of by weight (Punkt vor Strich)

brotkrueml: Hmm, then we should write something like "... instead of order of operations (multiplication and division before addition and subtraction)." What do you think?

linawolf: I updated the suggestion

linawolf (approved):

Issue #658, opened 2022-11-28
Replace ### BEGIN~OF~TABLE ### construct with custom confval directives


linawolf:
We want to replace all constructs to display properties with tables

i.e.

```
.. ### BEGIN~OF~TABLE ###

.. container:: table-row

   Property
         width

   Data type
         integer /:ref:`stdWrap <stdwrap>`

   Description
         Width of the SVG.

   Default
         600
```

with custom confvals that carry information about the type. The confval directive is then once defined in the `settings.cfg` :

```
[sphinx_object_types_to_add]

cobj-text =  cobj-text // cobj-text // Content object TEXT
cobj-svg =  cobj-svg// cobj-svg// Content object SVG
# ...
```

And then replaces the table rows:

```
..  cobj-svg:: width

    :Data type: integer /:ref:`stdWrap <stdwrap>`
    :Default: 600

    Width of the SVG.
```

These entries receive automatic unique links, they are displayed in the Index in their own Keyword etc.

PR #663, opened 2022-11-28
[TASK] Introduce custom confval directive for Data Processors

linawolf:
This also removes the warnings about duplicate confval names

references #658, https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/pull/310
releases: main, 11.5

### DIFF
--- a/Documentation/Settings.cfg
+++ b/Documentation/Settings.cfg
@@ -25,18 +25,18 @@ use_opensearch       =
 
 [sphinx_object_types_to_add]
 
-t3-cobj-case- =  t3-cobj-case- // t3-cobj-case- // Content object CASE
+t3-cobj-case- = t3-cobj-case- // t3-cobj-case- // Content object CASE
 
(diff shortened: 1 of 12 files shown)


Documentation/Settings.cfg
```diff
@@ -23,6 +23,19 @@ project_discussions  =
 
 use_opensearch       =
 
+[sphinx_object_types_to_add]
+
+data-processor-csv =  data-processor-csv // data-processor-csv // Data processor CommaSeparatedValueProcessor
```
brotkrueml, line 28: Can you add a comment what this syntax means?

linawolf: As this is a general new feature provided by the new rendering container it should be documented in How to document, not in each and every place we use it?

brotkrueml: I am fine, as above is the comment where to find that syntax (https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/GeneralConventions/FileStructure.html#settings-cfg). Can be add this information there first? Want to understand before approving the PRs.

marble (approved):

PR #689, opened 2023-01-24
Remove DC-tag example

julianhofmann:
Dublin Core tags are not supported since the introduction of the Meta Tag API. Neither the core itself covers DC tags, nor does EXT:seo.

### DIFF
--- a/Documentation/Setup/Page/Index.rst
+++ b/Documentation/Setup/Page/Index.rst
@@ -1028,7 +1028,6 @@ meta
                    og:site_name = TYPO3
                    og:site_name.attribute = property
                    description = Inspiring people to share Normal
-                   dc.description = Inspiring people to share [DC tags]
                    og:description = Inspiring people to share [OpenGraph]
                    og:description.attribute = property
                    og:locale = en_GB


brotkrueml:
> Dublin Core tags are not supported since the introduction of the Meta Tag API. Neither the core itself covers DC tags, nor does EXT:seo.

Are you sure? I thought, the page.meta things are taken just as is? I have to admit, I haven't looked into the code yet or tried it out.

julianhofmann:
Not sure enough... :-/ SEO is not my daily work. But today I tried to set some DC-tags an it would not work.

After some more googling, I've found a changelog entry, which helps me: [Feature: #67360 - Custom attribute name and multiple values for meta tags](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/7.4/Feature-67360-CustomAttributeNameAndMultipleValuesForMetaTags.html)

We've lost an escaping backslash in the docs somewhen:

❌ Current, wrong version:
```
page {
    meta {
        dc.description = Inspiring people to share [DC tags]
    }
}
```

 ✅ Correct one (with an escaped dot in the key)
page {
    meta {
        dc\.description = Inspiring people to share [DC tags]
    }
}

brotkrueml (approved):

PR #691, opened 2023-01-27
[DOCS] Hint to enable defaultJS to use spamProtect

helsner:


Documentation/Setup/Config/Index.rst
```diff
@@ -2053,6 +2053,8 @@ spamProtectEmailAddresses
          setting it to >= 2 means a "z" is converted to "\|" which is a special
          character in TYPO3 tables syntax – and that might confuse columns in
          tables.)
+         
+         It is required to enable the default JS and not disable it. (see `removeDefaultJS`)
```
brotkrueml, line 2057:
```suggestion
         It is required to enable the default JavaScript and not disable it. (see :ref:`removeDefaultJS <setup-config-removedefaultjs>`)
```

brotkrueml (approved):

PR #699, opened 2023-02-09
use [END] in upper case letters

franzholz:


brotkrueml (commented):
Reason?

franzholz:
Please: All examples should be equal.

linawolf:
The Core is predominantely using `[END]` in upper Case, however `[GLOBAL]` is much more common. I would opt for changing them all to upper case.

linawolf (changes requested):
Let us change all examples to upper case `[END]` as the Core does

brotkrueml (approved):

linawolf (approved):

linawolf:
@franzholz thanks for catching this!

PR #708, opened 2023-03-07
Mention returnLast option 'result'

kitzberger:
Introduce with release 11.4 via this commit: https://github.com/TYPO3/typo3/commit/9ccd931294196d18a34acc5e18eec8e167d65115

### DIFF
--- a/Documentation/Functions/Typolink.rst
+++ b/Documentation/Functions/Typolink.rst
@@ -427,6 +427,8 @@ returnLast
 
     So, in these two cases you will not get the value wrapped but the URL
     or target value returned!
+    
+    If set to ``result``, it will return the ``LinkResult`` object (only since TYPO3 11.4)
 
 section
 -------


linawolf:
Mhh, I believe TypoScript does not return PHP Objects. Have you tried what happens when you set this option?

The changelog entries are talking about the "AbstractTypolinkBuilder" but not about a new setting option in TypoScript

kitzberger:
@linawolf, correct me if I'm wrong but typolinks are not only used in TypoScript. And in TYPO3 11 this is returning the whole link result: https://github.com/TYPO3/typo3/blob/11.5/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L4980-L4988

I believe in 12 it's a string again?

brotkrueml (commented):
When using this TypoScript in v11:

    page.40 = TEXT
    page.40.typolink {
      parameter = 11
      returnLast = result
    }

I am getting this as an output in frontend:

```json
{
  "href": "\/my-page",
  "target": null,
  "class": null,
  "title": null,
  "linkText": "My page",
  "additionalAttributes": []
}
```

So, no `LinkResult` object but a representation as JSON string (from the `__toString()` method of the object). If it makes sense to use it? Well, don't know, but it is a valid option with an output and may be documented therefore ;-)

DavidBruchmann:
no experience with headless, but looks usable for it

brotkrueml:
> no experience with headless, but looks usable for it

But you won't build your headless API with TypoScript, I assume ;-)

kitzberger:
> But you won't build your headless API with TypoScript, I assume ;-)

It's possible though and that's exactly what EXT:headless is doing.

DavidBruchmann (changes requested):

Documentation/Functions/Typolink.rst
```diff
@@ -427,6 +427,8 @@ returnLast
 
     So, in these two cases you will not get the value wrapped but the URL
     or target value returned!
+    
+    If set to ``result``, it will return the ``LinkResult`` object (only since TYPO3 11.4)
```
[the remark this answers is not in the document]
kitzberger: @DavidBruchmann, @brotkrueml, I just added some more information but kept the notion of the `LinkResult` object. I find it kinda helpful to have that information at hand.

Documentation/Functions/Typolink.rst
```diff
@@ -428,7 +428,18 @@ returnLast
     So, in these two cases you will not get the value wrapped but the URL
     or target value returned!
     
-    If set to ``result``, it will return the ``LinkResult`` object (only since TYPO3 11.4)
+    If set to ``result``, it will return the json_encoded output of the ``LinkResult`` object (only since TYPO3 11.4)
```
DavidBruchmann, line 431: for me is the info in the round brackets not clear.
Does the parameter value `result` exist since TYPO3 v11.4 or was the output just differently before?

For the first case I propose this:
```
If set to ``result`` (available since TYPO3 11.4), it will return the json_encoded output of the ``LinkResult`` object.
```
Not related to this issue in particular: @linawolf is it possible that with changed rights I could propose with the code icon so that my propositions could be confirmed and applied with a single click?

DavidBruchmann (dismissed):

Documentation/Functions/Typolink.rst
brotkrueml: We should use the directive `.. versionadded:: 11.4`.

kitzberger: @brotkrueml, alright, think I got it now :muscle: Please squash it when merging ;)

Documentation/Functions/Typolink.rst
```diff
@@ -428,6 +428,21 @@ returnLast
     So, in these two cases you will not get the value wrapped but the URL
     or target value returned!
 
+    ..  versionadded:: 11.4
+        If set to ``result``, it will return the json_encoded output of the
```
brotkrueml, line 432: Can you remove one indentation for the text (and add an empty line)? We use `.. versionadded:: 11.4` just as a "header". Then it is fine for me.

So like 

    ..  versionadded:: 11.4

    If set to ....

kitzberger: @brotkrueml, sure thing. done.

You might wanna check other parts of the docs then, since I got that inline-ness from somewhere in the docs. E.g. in here
* `Documentation/Conditions/Index.rst` and
* `Documentation/ContentObjects/Fluidtemplate/DataProcessing/*.rst`

brotkrueml:
@kitzberger Thanks for the catch and your contribution :-)

PR #719, opened 2023-04-01
[DEPRECATION] #100349 - Deprecate loginUser() and usergroup() conditions

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/434
Releases: main

PR #720, opened 2023-04-01
[TASK] Add note to avoid loginUser and usergroup conditions

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/434
Releases: 11.5

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1261,6 +1261,14 @@ compatVersion
 loginUser
 ~~~~~~~~~
 
+..  note::
+    This method is deprecated in TYPO3 v12.4. The transition can be done in
+    existing TYPO3 v11 projects already. You should use either
+    :ref:`frontend.user <condition-frontend-user>` to test for frontend user
+    state (available in frontend TypoScript), or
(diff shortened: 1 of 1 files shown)


lolli42:
this will be handled by my bigger condition-refactoring patch in this area. maybe wait for its merge, then have a look at it again.

brotkrueml:
> this will be handled by my bigger condition-refactoring patch in this area. maybe wait for its merge, then have a look at it again.

So another patch will adjust the docs again. Okay, will wait ;-)

lolli42:
i think it's done and merged now in main. unsure if you still want something for v11.

brotkrueml:
> i think it's done and merged now in main. unsure if you still want something for v11.

It is only a hint for v11 users that they should use the frontend.user and backend.user conditions instead of the in v12 deprecated ones to stay forward-compatible. So they avoid additional work when upgrading.

So, yes, only v11 for this patch. v12 is another patch:
https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/pull/719

linawolf (approved):

Issue #712, opened 2023-03-13
breadcrumb example snippet renders with breadcrumb ending with">"


sypets:
### Current behavior

Using the snippet from https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/ContentObjects/Hmenu/Rootline.html#properties

```
page = PAGE
page.2 = HMENU
page.2.special = rootline
page.2.special.range = 1|-2
page.2.special.targets.3 = page
page.2.1 = TMENU
page.2.1.target = _top
page.2.1.wrap = <HR> | <HR>
page.2.1.NO {
  linkWrap = | >
}

```

results in a breadcrumb ending with ">" which is probably not the desired outcome

e.g.

Example breadcrumb:

> test > typoscript >

### Expected behavior/output

An example without trailing ">". I would expect the ">" to only be _between_ the elemens.

### Possible Solution

Would provide example, but we handle that with css, not TypoScript. Also, the various wrap thingies not so well documented, so kind of clueless here.

## Links

https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/MenuObjects/Tmenuitem/Index.html#linkwrap

brotkrueml:
I assume, then optionSplit comes into place. Personally I would set the `>` via CSS and use `<ol><li>...</li></ol>`. I also don't understand why `<HR>` is used here for wrapping an item?

DavidBruchmann:
The wrap thingies are documented on https://docs.typo3.org/m/typo3/reference-typoscript/11.5/en-us/Functions/OptionSplit.html and even one detail is missing there (noting the matches themselves with a wrap, like &lt;span&gt;|&lt;/span&gt;) it's quite extensive.  
Nevertheless the page shows that the result depends on the depth respectively amount of elements and there can't be a universal solution how it seems to be required for breadcrumb menus. It gets still much more complicated if the first, last / current and middle position shall be recognized to apply classes. 
So I propose to change the example and just make a list like you mentioned.  
The <hr> can be removed, it doesn't add any value, in my opinion, but rather might be confusing.

PR #722, opened 2023-04-02
[TASK] Adjust rootline example to use a HTML list

brotkrueml:
Additionally:
- Add links
- Use code directives

Resolves: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/issues/712
Releases: main, 11.5

### DIFF
--- a/Documentation/ContentObjects/Hmenu/Rootline.rst
+++ b/Documentation/ContentObjects/Hmenu/Rootline.rst
@@ -2,9 +2,9 @@
 .. index:: HMENU; special = rootline
 .. _hmenu-special-rootline:
 
-===========================
+==========================
 Rootline - breadcrumb menu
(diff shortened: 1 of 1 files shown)


DavidBruchmann:
What's about adding another wrap:
```
page.2.wrap = You're here: |
```

brotkrueml:
> What's about adding another wrap:
> 
> ```
> page.2.wrap = You're here: |
> ```

We should not over-engineer this example. That is part of a designer/UX specialist to define the requirements for a breadcrumb on a website. We should stick to the purpose:

    Here the links to pages on level 3 will have :html:`target="page"`, while
    all other levels will have :html:`target="_top"`

DavidBruchmann (approved):

PR #730, opened 2023-04-14
Move example with typeNum

sfroemkenjw:
Hello,

with following change:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.4/Deprecation-100405-PropertyTypoScriptFrontendController-type.html

you can not access $TSFE->type anymore. That's why I have moved the example to the new request.getPageAttributes() location.

But be careful:
Doing so, will result in an Exception while opening the ExtensionManager. I already have added a forge ticket: https://forge.typo3.org/issues/100595

Stefan

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -884,9 +884,6 @@ getTSFE()
    .. code-block:: typoscript
       :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
 
-      # True if current typenum is 98
-      [getTSFE().type == 98]
-
       # True if current page uid is 17. Use the page variable instead
       [getTSFE().id == 17]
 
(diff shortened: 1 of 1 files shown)


brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/444

brotkrueml:
@sfroemkenjw Thank you for the PR. I suggest to wait with the merge until the mentioned forge issue is resolved.

brotkrueml (changes requested):
The issue in backend seems to be fixed in main branch (according to Oli and also I cannot reproduce it anymore).

But: The section you adjusted is about `getTSFE()`. Therefore, I suggest to remove the `getTSFE().type`  example there. Additionally, add the typenum snippet in the `request.getPageArguments()` section as example:
https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/Conditions/Index.html#request-getpagearguments

sfroemkenjw:
Please have a look into my change. It exactly does what you describe ;-)

brotkrueml (approved):

PR #736, opened 2023-04-25
only tt_content is used as standard

franzholz:
Maybe it has been intended to use tt_* tables for content rendering 20 years ago. But no other table than tt_content has been used.  Other tables like tt_address have to be defined, before they can be used.

https://stackoverflow.com/questions/52973130/display-tt-address-records-when-inserted-via-insert-record-in-typo3-8-7

Other table names not starting with tt* could be used as well.

### DIFF
--- a/Documentation/TopLevelObjects/Other.rst
+++ b/Documentation/TopLevelObjects/Other.rst
@@ -37,12 +37,14 @@ temp, styles & lib
 
 .. index:: Top-level objects; tt_content
 .. _tlo-tt:
+.. _tlo-tt_content:
 
-====
-tt_*
-====
 
-``tt_``, e.g. ``tt_content`` (from “content (default)”) is used to render content from tables.
+==========
+tt_content
+==========
+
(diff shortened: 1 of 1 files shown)


Documentation/TopLevelObjects/Other.rst
```diff
@@ -37,12 +37,14 @@ temp, styles & lib
...
-====
-tt_*
-====
 
-``tt_``, e.g. ``tt_content`` (from “content (default)”) is used to render content from tables.
+==========
+tt_content
+==========
+
+``tt_content``,  (from “content (default)”) is used to render content from tt_content or other tables
```
linawolf, line 47:
```suggestion
The top level keyword :typoscript:`tt_content`, is used to render content from the table :sql:`tt_content`
```

linawolf:
I believe, you are right. Also the usage of the `tt_` prefix is being phased out. It is only there fore historic reasons in tt_content

linawolf (approved):

brotkrueml:
Should we backport this also to 11.5?

PR #746, opened 2023-06-13
Use hint to hint at HMENU options

mbrodala:
Attention should be used for caveats where things can go wrong.


brotkrueml (approved):

brotkrueml:
@mbrodala Thanks for your contribution :-)

PR #757, opened 2023-06-18
[TASK] #101053 - Remove deprecated code from ExpressionLanguage

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/534
Releases: main

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -795,73 +795,6 @@ compatVersion()
       [compatVersion("11.5.1")]
 
 
-.. index:: Conditions; loginUser
-.. _condition-function-loginUser:
-
-loginUser()
-===========
-
-..  deprecated:: 12.4
-    This function has been marked as deprecated with TYPO3 v12 and should not be
-    used anymore. Use the variables :typoscript:`frontend.user` and
-    :typoscript:`backend.user` instead. See the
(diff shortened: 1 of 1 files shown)


Documentation/Conditions/Index.rst
```diff
@@ -795,73 +795,6 @@ compatVersion()
       [compatVersion("11.5.1")]
 
 
-.. index:: Conditions; loginUser
-.. _condition-function-loginUser:
-
-loginUser()
-===========
-
-..  deprecated:: 12.4
```
linawolf, line 804: keep a hint or migration information maybe?

brotkrueml: I would remove all that old stuff.

linawolf: Migrating Conditions has caused a lot of trouble for users in the past

brotkrueml: Okay okay, will leave it in somehow

linawolf: We could maybe also make a "Migration" section, we had that in the past

brotkrueml: It is already in place in v12. Will just adjust the note above and leave all the stuff the same.

linawolf (approved):

PR #758, opened 2023-06-19
[TASK] #98200 Recommend conditions for rootPaths

s2b:
As stated by Oliver Hader in the [forge issue](https://forge.typo3.org/issues/98200):

> Using stdWrap on declarative settings like paths, which are processed by other components (e.g. Fluid), is not a good idea...


brotkrueml (approved):

brotkrueml:
@s2b Thanks for the improvement.

PR #761, opened 2023-06-27
Remove .dataProcessing from FilesProcessor docs

dablach:
I'm not quite sure here. But if I do understand the code correctly, the .dataProcessing subkey inside a FilesProcessor configuration is not interpreted (unlike DatabaseQueryProcessor). It wouldn't even be clear, how the resulting data would get passed alongside the fal objects.


dablach:
I did check the code. But I'm not quite sure if I understood it correctly. I would appreciate if someone else could confirm before merging this.

linawolf (approved):
I think, you are right here

PR #770, opened 2023-07-31
Add example for numRows

kitzberger:


brotkrueml (commented):
@kitzberger Thanks for your contribution. I suggest to split the one example into two examples - perhaps with explaining what they do with a small introduction sentence:
- "Get the total number of content elements available"
- "Get the number of all content elements in col pos = 1"

As `numRows` is a function the `page.10.variables` is not really necessary in the example. I assume, `page.10` is a `FLUIDTEMPLATE`? If you want to keep it, please also add the FLUIDTEMPLATE so the reader can grasp the context, like:

    10 = FLUIDTEMPLATE
    10 {
        # ... your other configuration
        variables {
            # your examples
        }
    }

What do you think?

kitzberger:
> What do you think?

Yeah, you're right! I've added a little description for the "more complex" snippet and removed the simpler one.

brotkrueml (approved):

PR #775, opened 2023-08-10
Add example for CONTENT select.join

sypets:
- Add small example for select.join.
- Improve explanatory text

Previously, the explanatory text was slightly misleading because it instructed to use the table name where in fact an entire join clause (wihtout the word "JOIN") can be used including aliasing the table and adding a constraint with "ON".


Documentation/Functions/Select.rst
```diff
@@ -267,9 +269,29 @@ join, leftjoin, rightjoin
 
     :Data type: :t3-data-type:`string` / :ref:`stdWrap`
 
-    Enter the table name for JOIN, LEFT OUTER JOIN and RIGHT OUTER JOIN
+    Enter the JOIN clause without for JOIN, LEFT OUTER JOIN and RIGHT OUTER JOIN
```
brotkrueml, line 272:
```suggestion
    Enter the JOIN clause without :sql:`JOIN`, :sql:`LEFT OUTER JOIN` and :sql:`RIGHT OUTER JOIN`
```

I assume, the "for" I removed is superfluous?

brotkrueml (approved):

brotkrueml:
@sypets Thanks for the improvement :+1:

PR #776, opened 2023-08-10
Optimize the links on select page

sypets:
- replace linking to "select" (the same) page with link to CONTENT object type
- point out additional examples on CONTENT page because this contains several examples with select and more context

### DIFF
--- a/Documentation/Functions/Select.rst
+++ b/Documentation/Functions/Select.rst
@@ -403,5 +403,6 @@ Condensed form:
        }
     }
 
+For more complete examples see :ref:`cobj-content`.
 
-See also: :ref:`if`, :ref:`select`, :t3-data-type:`wrap`, :ref:`stdWrap`, :ref:`data-type-cobject`
+See also: :ref:`if`, :ref:`cobj-content`, :t3-data-type:`wrap`, :ref:`stdWrap`, :ref:`data-type-cobject`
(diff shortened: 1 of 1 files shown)


Documentation/Functions/Select.rst
```diff
@@ -403,5 +403,6 @@ Condensed form:
        }
     }
 
+For more complete examples see :ref:`cobj-content`.
 
-See also: :ref:`if`, :ref:`select`, :t3-data-type:`wrap`, :ref:`stdWrap`, :ref:`data-type-cobject`
+See also: :ref:`if`, :ref:`cobj-content`, :t3-data-type:`wrap`, :ref:`stdWrap`, :ref:`data-type-cobject`
```
brotkrueml, line 408: The second link to cobj-content is superfluous, as the paragraph above links cobj-content?

The last link goes to data-type-cobject. At first glance, I don't know why? Has this a value for the user?

sypets: I removed the the link to data-type-cobject, to if and the duplicate cobj-content as you suggested.
I created a list with some more explanatory text so context is more clear when clicking on a link and how this might be relevant.

To be honest I wasn't sure if any of the other links are helpful or relevant. I don't think they were really relevant for the select itself.

brotkrueml (approved):

brotkrueml:
@sypets Thanks for your contribution :+1:

PR #793, opened 2023-09-07
Changing the order of the Content-supplying properties of stdWrap part

FlorianT03:
releases: main

Remark: In my opinion the "Example" and "Content-supplying properties" part, at the end of the site, should be moved back to the top, like in for example the 10.x version of the TYPO3 Documentation. Because if it is so far down, hardly anyone reads it, but especially the part with the order is very important. If you see this differently or have suggestions for improvement through mistakes, then let me know.

https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/Functions/Stdwrap.html

### DIFF
--- a/Documentation/Functions/Stdwrap.rst
+++ b/Documentation/Functions/Stdwrap.rst
@@ -14,26 +14,8 @@ When a data type is set to "*type* /stdWrap" it means that the value
 is parsed through the stdWrap function with the properties of the
 value as parameters.
 
-.. _stdwrap-examples:
-
-Example
-=======
-
-Example with the property "value" of the content object ":ref:`cobj-text`":
-
-.. code-block:: typoscript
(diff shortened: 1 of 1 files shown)


brotkrueml:
IMHO, I agree with the content-supplying part (as the ordering may be important dependend on the use case) but not with the example. Mostly, the examples are at the end of the page and this example makes no difference to others, it is just a simple example. We should not move the TOC too far down.

FlorianT03:
@brotkrueml well thanks for your honest opinion! tbh at first I was not even switching the place of the example, should have left it like that. I will change it, when I have time again^^

brotkrueml:
@linawolf I disagree, but feel free to approve.

linawolf:
@brotkrueml ok, lets stick to status quo for now

linawolf (approved):

PR #795, opened 2023-09-11
Fix wrong typolink syntax

kitzberger:
The section parameter for typolink is not prefixed with a "c" in the database, that "c" is added during FE rendering.


brotkrueml (approved):

brotkrueml:
@kitzberger Good catch! Thanks for your contribution :+1:

PR #800, opened 2023-09-19
[TASK] Use current TYPO3 logo for gifbuilder examples

leagt:
Switch from old orange-green logo to the all orange one.


brotkrueml:
FYI: This is the page with the images:
https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/Gifbuilder/Examples.html

linawolf (approved):
Thanks for the new screenshots!

brotkrueml (approved):

PR #803, opened 2023-09-22
[TASK] #101941 - Remove GIFBUILDER's reduceColors option

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/653
Releases: main

### DIFF
--- a/Documentation/Gifbuilder/ObjectNames/Index.rst
+++ b/Documentation/Gifbuilder/ObjectNames/Index.rst
@@ -140,19 +140,6 @@ different each time!
          "jpg"/"jpeg" = jpg-image
 
 
-
-.. container:: table-row
-
-   Property
-         reduceColors
-
-   Data type
-         positive integer (1-255) /:ref:`stdWrap <stdwrap>`
-
-   Description
-         Reduce the number of colors.
-
-
 .. container:: table-row
 
    Property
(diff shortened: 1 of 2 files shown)


sarahmccarthy123 (approved):
Looks fine

Documentation/Gifbuilder/ObjectNames/Index.rst
```diff
@@ -197,8 +184,8 @@ different each time!
          This will allow for the closest color to be matched instead. You may
          need this if you image is not guaranteed "clean".
 
-         **Note:** You may experience that this does not work, if you use the
-         reduceColors-option or render text with niceText-option.
+         **Note:** You may experience that this does not work, if you render
+         text with the niceText option.
```
linawolf, line 188:
```suggestion
         text with the :typoscript:`niceText` option.
```
I think this is a TS option?

brotkrueml: I assume so, I would like to link that (if it is an option, but this would be another PR). But the Gifbuilder docs need more love anyway.

linawolf: Ok, lets do it in a follow up then

linawolf (approved):

PR #805, opened 2023-09-28
Fix links

sypets:
https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/Functions/Typolink.html


brotkrueml (approved):

brotkrueml:
@sypets Good catch! Thanks for your contribution :+1:

PR #813, opened 2023-10-23
Add info that data.path resolves EXT: syntax

DanielSiepmann:


linawolf (approved):

brotkrueml (approved):

brotkrueml:
@DanielSiepmann Thanks for the addition :+1:

DanielSiepmann:
Wasn't sure whether it is too small. But I didn't like the idea to look it up over and over again in the future.
And this would probably be the case with v12 and the new filesystem setup. Glad you like it.

brotkrueml:
It is an important piece of information - and it doesn't matter, if it is just some words or a whole section.

PR #816, opened 2023-10-27
[TASK] Revise Gibuilder/Gifbuilder chapter

brotkrueml:
Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/Gifbuilder/Gifbuilder/Index.rst
+++ b/Documentation/Gifbuilder/Gifbuilder/Index.rst
@@ -7,7 +7,7 @@ GIFBUILDER
 ==========
 
 GIFBUILDER is an object type, which is used in many situations for creating
-image files (for example, GIF, PNG or JPG). Wherever the GIFBUILDER object
+image files (for example, GIF, PNG or JPG). Wherever the ->GIFBUILDER object
 type is mentioned, these are the properties that apply.
 
 Using TypoScript, you can define a "numerical array" of
--- a/Documentation/Gifbuilder/Gifbuilder/Index.rst
+++ b/Documentation/Gifbuilder/Gifbuilder/Index.rst
@@ -1,22 +1,22 @@
-.. include:: /Includes.rst.txt
-.. index:: GIFBUILDER
-.. _gifbuilder:
+..  include:: /Includes.rst.txt
+..  index:: GIFBUILDER
+..  _gifbuilder:
 
 ==========
 GIFBUILDER
 ==========
 
 GIFBUILDER is an object type, which is used in many situations for creating
-image files (e.g. gif, png or jpg). Anywhere the ->GIFBUILDER object type
-is mentioned, these are the properties that apply.
+image files (for example, GIF, PNG or JPG). Wherever the GIFBUILDER object
+type is mentioned, these are the properties that apply.
 
-Using TypoScript you can define a "numerical array" of "GIFBUILDER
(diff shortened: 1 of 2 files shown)


Documentation/Gifbuilder/Gifbuilder/Index.rst
```diff
@@ -1,73 +1,75 @@
...
-.. _gifbuilder:
+..  include:: /Includes.rst.txt
+..  index:: GIFBUILDER
+..  _gifbuilder:
 
 ==========
 GIFBUILDER
 ==========
 
 GIFBUILDER is an object type, which is used in many situations for creating
```
linawolf, line 9:
```suggestion
:typoscript:`GIFBUILDER` is an object type, which is used in many situations for creating
```
Should we mark gifbuilder as TypoScript throughout as that is a word from TypoScript and not from english?

Documentation/Gifbuilder/Gifbuilder/Index.rst
```diff
@@ -1,73 +1,75 @@
...
+
+On using the special function :typoscript:`max()`, the maximum of multiple
+values can be determined. Example:
+
+..  code-block:: typoscript
+    :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
+
+    XY = [10.w]+[20.w], max([10.h], [20.h])
+
+Here's a full example (taken from "styles.content (default)" of older TYPO3
```
linawolf, line 47: Not sure how helpful this example is?

brotkrueml: At least it is an example. I never worked with GIFBUILDER in the past, so have not really a clue what is going on here. The whole working from me on GUIFBUILDER now is preparing it for adding a changelog entry for v13 (support of webp).

The example can be tackled in a separate PR.

Documentation/Gifbuilder/Gifbuilder/Index.rst
brotkrueml: Thought also about it, but there are different mentions of GIFBUILDER, sometimes referencing as "GIFBUILDER object". At first glance, I don't know if it is referring to the TypoScript cObject or something else. So I stpped away from marking all the GIFBUILDER occurrences as TypoScript for now.

Documentation/Gifbuilder/Gifbuilder/Index.rst
linawolf: ack

Documentation/Gifbuilder/Gifbuilder/Index.rst
linawolf: I think it is the cObject, all cObjects are classes under the hood and therefore can also be called object

brotkrueml: @linawolf Do you know, why sometimes `GIFBUILDER` is used, and once `->GIFBUILDER`? Is it the same? I also saw this with, for example, `->if`  and `if`.

linawolf: I believe one referes to the TYPOSCRIPT Object and one to the corresponding function in PHP? But maybe also no one thought about it?

brotkrueml: Don't know, leave it as is for now.

linawolf: Ack

linawolf (approved):

PR #853, opened 2023-10-31
[TASK] Dissolve Gifbuilder/Gifbuilder chapter

brotkrueml:
Under the Gifbuilder folder there is another Gifbuilder folder with the introduction text. The first Gifbuilder folder holds only a TOC.

This patch changes that:
- The introduction text is moved to the start page.
- The text about +calc is moved parallel to the start page. This way it is more prominent as it is displayed in the navigation.
- The TOC now starts with the examples and the +calc chapter. So they are more prominent in the TOC. Then the object reference is following.

The link to the NonGifbuilderObject is a left-over: It was removed in a previous patch and forgotten to remove the link.

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/Gifbuilder/Calc.rst
+++ b/Documentation/Gifbuilder/Calc.rst
@@ -1,25 +1,5 @@
 ..  include:: /Includes.rst.txt
-..  index:: GIFBUILDER
-..  _gifbuilder:
-
-==========
-GIFBUILDER
-==========
-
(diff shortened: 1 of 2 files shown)


brotkrueml:
I will also take care of the redirects.

linawolf (approved):

linawolf:
You are doing really great improvements to this chapter! I love it!

PR #859, opened 2023-11-01
[TASK] Move GIFBUILDER Properties page

brotkrueml:
The "Properties" page is no GIFBUILDER object, therefore it is moved outside the "ObjectNames" folder and is now a subpage of the main GIFBUILDER folder. This way it is also more prominent in the navigation.

Releases: main, 12.4, 11.5


linawolf (approved):

linawolf:
Will you make a redirect for that?

PR #881, opened 2023-11-06
FIX: wrong level of 'sword' in TOC

kitzberger:
"sword" is rendered wrong in the TOC here: https://docs.typo3.org/m/typo3/reference-typoscript/11.5/en-us/Functions/Parsefunc.html


brotkrueml (approved):

brotkrueml:
@kitzberger Thanks for catching this :+1:

PR #900, opened 2023-11-12
[TASK] Use custom types for linking object properties in GIFBUILDER chapter

brotkrueml:
Releases: main, 12.4, 11.5


linawolf (commented):
So would you like to continue to replace all the former simple references with custom once? I thought we wanted to put custom anchors and link them?

brotkrueml:
> So would you like to continue to replace all the former simple references with custom once? I thought we wanted to put custom anchors and link them?

I don't know yet, but as you started with them last year and I started also with the GIFBUILDER without thinking really about it, well ...

brotkrueml:
And you've got a valid point, a separate type has better semantics than using a confval directive.

linawolf (approved):

PR #904, opened 2023-11-14
Add label for addPageCacheTags again

sypets:
A label existed previously (stdwrap-addpagecachetags) and is still being referenced from TSconfig reference (see warnings).


brotkrueml (approved):

brotkrueml:
@sypets Thanks :+1:

PR #910, opened 2023-11-21
Fix variable name for userGroupIds

RTPAst:


brotkrueml (approved):

brotkrueml:
@RTPAst Good catch! Thanks for the fix :+1:

PR #925, opened 2023-12-09
[BUGFIX] Reintroduce label

linawolf:
As it is referenced from the changelog

releases: main, 12.4


brotkrueml (approved):

brotkrueml:
maybe also 11.5 as it does not hurt?

linawolf:
Trying, but I think it was only removed in 12

PR #953, opened 2023-12-26
[TASK] #102715 - Add alternatives for TSFE properties

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/770
Releases: main


linawolf (approved):
I think this is very helpful, thanks for putting work into this

PR #1007, opened 2024-01-20
[BUGFIX] Typo 'best practiSe'

lolli42:


brotkrueml:
Actually, both is valid: https://www.merriam-webster.com/dictionary/practise

But the "c" variant is more common.

@lolli42 Thanks for the PR.

brotkrueml (approved):

PR #1011, opened 2024-01-20
[TASK] Avoid 'types' TLO

lolli42:
This one is declared 'internal' in the docs
already and may be removed in v13. There is
no point in documenting this core internal
structure. The 'plaintext' hint is wrong.


brotkrueml (approved):

brotkrueml:
@lolli42 Thanks for the cleanup.

PR #1066, opened 2024-02-07
[TASK] Resolve todos in HtmlparserTags chapter

brotkrueml:
See: https://github.com/TYPO3/typo3/blob/12.4/typo3/sysext/core/Classes/Html/HtmlParser.php#L512-L517
Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/Functions/HtmlparserTags.rst
+++ b/Documentation/Functions/HtmlparserTags.rst
@@ -108,10 +108,8 @@ fixAttrib.[attribute].upper
 
     :Data type: :ref:`data-type-boolean`
 
-    If true, the value is passed through the
-    respective PHP-function.
-
-    .. TODO: There is not such PHP function, probably strtoupper?
+    If true, the value is passed through the PHP function
+    `strtoupper() <https://www.php.net/manual/en/function.strtoupper.php>`__.
 
 fixAttrib.[attribute].lower
 ---------------------------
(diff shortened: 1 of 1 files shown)


Documentation/Functions/HtmlparserTags.rst
```diff
@@ -108,10 +108,8 @@ fixAttrib.[attribute].upper
 
     :Data type: :ref:`data-type-boolean`
 
-    If true, the value is passed through the
-    respective PHP-function.
-
-    .. TODO: There is not such PHP function, probably strtoupper?
+    If true, the value is passed through the PHP function
+    `strtoupper() <https://www.php.net/manual/en/function.strtoupper.php>`__.
```
garvinhicking, line 112: Wonder if we might link to mb_stroupper instead? My guess is that function is used instead of the multibyte-unaware one...? I guess that's the reason why it was written as "respective PHP-function", because it depends on the existence of the php-mbstring extension (which is recommended)...?

brotkrueml: Look at the link I provided in the commit message: Core uses strtoupper() and strtolower(), so this is the way it (currently) works.

garvinhicking: True. I was completely sure that the core used mb_* functions through a layer like GeneralUtility. Thanks for clearing this up for me.

garvinhicking (approved):

PR #1080, opened 2024-02-29
[TASK] Replace "t3-function-htmlparser" with "confval"

brotkrueml:
This is a preparation for switching to PHP-based documentation rendering.

Additionally:
- Anchors are added

Releases: main, 12.4


linawolf (approved):
We are getting there!!

PR #1113, opened 2024-03-04
[BUGFIX] Remove noopener as superfluous

brotkrueml:
Initially this feature added `rel="noopener noreferrer`. However `noreferrer` also implies the property `noopener`. Therefore, the later was removed.

Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/867
Releases: main, 12.4, 11.5


linawolf (approved):
Good catch!

PR #1118, opened 2024-03-05
[TASK] Replace "t3-function-stdwrap" with "confval"

brotkrueml:
This is a preparation for switching to PHP-based documentation rendering.

Additionally:
- Anchors are added
- Use "t3src" to link directly to a Core file
- Adjust header level according to https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/CheatSheet.html#headlines

Releases: main, 12.4


brotkrueml:
Only two to go then :rocket:

PR #1133, opened 2024-03-24
[TASK] Add ddev directory to gitignore

linawolf:
Most recent versions of ddev seem to not create their own gitignore anymore

releases: main, 12.4, 11.5


garvinhicking (approved):
Odd, mine still does. But I think its ok to add it anyways ;)

PR #1140, opened 2024-03-24
[TASK] Set unique names for confvals in datatypes chapter

linawolf:
releases: main, 12.4


brotkrueml (approved):
This should also be backported to 11.5 to avoid rendering errors there.

PR #1143, opened 2024-03-24
[TASK] Set unique names for confvals in conditions chapter

linawolf:
releases: main, 12.4

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -652,7 +652,7 @@ date()
 ======
 
 ..  confval:: date()
-    :name: condition-function-date
+    :name: condition-date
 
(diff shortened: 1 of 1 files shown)


Documentation/Conditions/Index.rst
```diff
@@ -1039,6 +1083,7 @@ session()
 =========
 
 ..  confval:: session()
+    :name: condition-functions-in-frontend-context-function-session
```
brotkrueml, line 1086: Why this long name? Also some other places.

brotkrueml (commented):
Should also be backported to 11.5 to avoid errors on rendering.

brotkrueml (approved):

PR #1162, opened 2024-03-26
[TASK] Unify confvals for config section

linawolf:
Replace aspect table with confval where not yet happened.

Use unified names

We should have a look at the default values in a follow up. Some are missing, some are integers for type boolean.

Releases: main, 12.4

### DIFF
--- a/Documentation/Setup/Config/Index.rst
+++ b/Documentation/Setup/Config/Index.rst
@@ -45,17 +45,13 @@ Properties of 'config'
     :local:
 
 
-
-..  ### BEGIN~OF~TABLE ###
-
-
 ..  _setup-config-absrefprefix:
 
 absRefPrefix
 ------------
 
 ..  confval:: absRefPrefix
-    :name: setup-config-absrefprefix
+    :name: config-absrefprefix
 
     :Data type: :ref:`data-type-string`
     :Special value: "auto"
@@ -110,7 +106,7 @@ additionalHeaders
 -----------------
 
 ..  confval:: additionalHeaders
-    :name: setup-config-additionalheaders
+    :name: config-additionalheaders
 
     :Data type: numerically indexed array of "HTTP header entries".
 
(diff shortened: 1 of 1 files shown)


garvinhicking:
Phew, this is hard to review, I could offer a vote of confidence only 😜

Documentation/Setup/Config/Index.rst
```diff
@@ -437,91 +397,78 @@ compressJs
...
 
-   Example
-         .. code-block:: typoscript
-            :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
+    config.concatenateJs = 1
 
-            config.concatenateJs = 1
+    page = PAGE
+    page.includeJSFooter {
+    test = fileadmin/user_upload/test.js
```
brotkrueml, line 466: Please indent correctly.

Documentation/Setup/Config/Index.rst
```diff
@@ -1568,26 +1329,22 @@ noPageTitle
...
+~~~~~~~
 
-   Example
-         .. code-block:: typoscript
-            :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
+..  code-block:: typoscript
+    :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
 
-            pageRendererTemplateFile = fileadmin/test_pagerender.html
+    pageRendererTemplateFile = EXT:my_extension/Ressources/Public/Templates/TestPagerender.html
```
brotkrueml, line 1347:
```suggestion
    pageRendererTemplateFile = EXT:my_extension/Ressources/Private/Templates/TestPagerender.html
```

This should not be exposed to the outside as this is used internally by PageRenderer.

Documentation/Setup/Config/Index.rst
linawolf: Done

brotkrueml (approved):
Reviewed it and it is fine for me, although haven't tested it. So, go ahead.

PR #1197, opened 2024-04-05
Set all other permalinks in Option Split file

Alagts:


linawolf:
Thanks for your changes, as anchors have to be unique Document-wide, I would prefix them all with "optionsplit-"

linawolf (approved):

PR #1208, opened 2024-04-08
[TASK] Overhaul appendix page

linawolf:
* Use confvals
* Correct indentation

releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/AppendixA/Index.rst
+++ b/Documentation/AppendixA/Index.rst
@@ -1,11 +1,11 @@
-.. include:: /Includes.rst.txt
-.. _appendix-a:
+..  include:: /Includes.rst.txt
+..  _appendix-a:
 
 ================================
 Appendix A – PHP include scripts
 ================================
 
-.. _appendix-include-scripts:
+..  _appendix-include-scripts:
 
 Including your script
 =====================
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
Actually, I don't understand what this section is about? The included script is part of the ContentObjectRenderer?

When I guess, I assume, this is about a USER function? But then the heading is totally misleading.

linawolf:
I agree the page needs a overhault, but not in this ticket...

Documentation/AppendixA/Index.rst
```diff
@@ -65,45 +65,48 @@ Call the function :php:`$GLOBALS['TSFE']->set_no_cache()`, if you want to
 disable caching of the page. Call this during development only! And
 call it, if the content you create may not be cached.
 
-**Note:** If you make a syntax error in your script that keeps PHP
-from executing it, then the :php:`$GLOBALS['TSFE']->set_no_cache()`
-function is not executed and the page *is* cached! So in these
-situations, correct the error, clear the page-cache and try again.
-This is true only for :typoscript:`USER` and not for :typoscript:`USER_INT`, which is
-rendered *after* the cached page!
+.. note::
```
brotkrueml, line 68:
```suggestion
..  note::
```

brotkrueml (approved):
Well, you named it "Overhaul" - that includes also the content IMHO ;-)
You sure, you want to do this? You'll have fun while backporting - and it's not the work worth IMHO.

PR #1209, opened 2024-04-09
Removed unnecessary new line

PKuhlmay:
This removes a blank line which caused a wrong output in the documentation. Without this blank line the text should not go into a new block.


brotkrueml (approved):
@PKuhlmay Thanks for the fix :+1:

PR #1214, opened 2024-04-09
[TASK] Restrict depth of TOC in data chapter

brotkrueml:
This removes the "Example:" headings from the TOC for better readability of the data keys. This also shortens the menu.

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/Functions/Data.rst
+++ b/Documentation/Functions/Data.rst
@@ -53,6 +53,7 @@ Properties
 
 ..  contents::
     :local:
+    :depth: 2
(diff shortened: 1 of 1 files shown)


brotkrueml:
Before:
before

After:
after

linawolf:
My idea for giving the examples speaking titles was actually so that you have an example for each data type as the names alone are sometimes not so clear what they do. Therefore I think this information is valuable in the list for those who do not know what a certain property is capabale of doing.

PR #1215, opened 2024-04-11
Update Index.rst

SZChimp:
Use the same Markup syntax for "defer" property in `includeJS` paragraph as for other option flags. Previously it was easily overlooked, because the "defer" option was not printed bold and its description was not in a newline below the bold text.


brotkrueml (approved):
@SZChimp Good catch :+1: Thanks for the fix.

brotkrueml:
@SZChimp Please make the PR against main in the future, this eases the ports to other branches/versions.

PR #1216, opened 2024-04-11
Fix markup for defer (#1215)

brotkrueml:
Use the same Markup syntax for "defer" property in `includeJS` paragraph as for other option flags. Previously it was easily overlooked, because the "defer" option was not printed bold and its description was not in a newline below the bold text.


brotkrueml:
Forward-port of #1215

linawolf (approved):

PR #1230, opened 2024-04-23
[TASK] Move the TMENU as subpage of the HMENU

linawolf:
TMENUs are seldom used anymore since there data processors and Templates are recommended. If they are used they are always used in the context of a HMENU. So the current situation where there is a menu point called "MENU" in the menu that doesnt tell you how to build a modern menu is quite misleading.

No text changes made, so that Git can understand the files where moved.

Text changes will be applied in a follow-up to further distinguish the usage.

I will create redirects

Releases: main, 12.4, 11.5


brotkrueml (commented):
I suggest to move TMENU and TMENUITEM directly under cObject and not to cObjects/HMENU, as this enhances the visibility in the main menu.

Additionally, in the main menu it is listed together with the HMENU properties (when moved under HMENU) which is sub-optimal IMHO and mixes up different things.

linawolf:
However the TMENU is not a content object and it cannot be used standalone

brotkrueml:
> However the TMENU is not a content object and it cannot be used standalone

https://github.com/TYPO3/typo3/blob/main/typo3/sysext/frontend/Classes/ContentObject/Menu/TextMenuContentObject.php

-> TextMenuContentObject

linawolf:
It extends the AbstractMenuContentObject  but not the AbstractContentObject. It cannot be used standalone and it is not an AbstractContentObject like HMENU or FLUIDTEMPLATE

brotkrueml:
Convinced, but will not have another look today.

PR #1242, opened 2024-05-11
[TASK] PAGEVIEW with confval tables

linawolf:
I decided to reorder the examples to a dedicated examples section. From each property / variable I link to the example. The advantage is, that we can use the same example for several properties or the same property with several examples.

Also the properties section is less cluttered.

image

### DIFF
--- a/Documentation/ContentObjects/Pageview/Index.rst
+++ b/Documentation/ContentObjects/Pageview/Index.rst
@@ -24,7 +24,7 @@ structure outlined below.
 
 ..  contents:: Table of contents
     :local:
-    :depth: 1
+    :depth: 2
 
 ..  _cobj-pageview-properties-example:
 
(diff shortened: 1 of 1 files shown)


linawolf:
If we would list the examples in the table I think it would be too cluttered:

image

brotkrueml (commented):
Currently, I am - let's say - reserved about this feature.

Personally, I like to see all the content at once, so I can scroll through the page and search in the page via browser). What I don't like are click orgys to display the content.

So, what problem do you want to solve with this feature? Then we can think about a solution.

linawolf:
@brotkrueml I think you misunderstood something, there is no click orgy or accordion effect here, that was a different discussion.

If you want to try it out please check out this PR and render it locally.

garvinhicking:
Haven't tried this out yet, and as long as it's not accordioned away I like having a linklist in the overview, good idea!

How about adding a ">> overview" link for each convfal that leads back to the overview? So one could better jump between a confval and the overview without scrolling, that could go well hand in hand?

linawolf:
Yes the links are navigational links just scrolling to the confval in question. A back to navigation link might be a good idea, that can be solved in rendering however as the directive I introduce here gives us full controll on how to display such things without needing to do rst-changes

brotkrueml:
> @brotkrueml I think you misunderstood something, there is no click orgy or accordion effect here, that was a different discussion.
> 
> If you want to try it out please check out this PR and render it locally.

I see an example link on the screenshot, with a click I assume to jump elsewhere on the page. To get back, I have to click on "back" of the browser. So, the content for a setting seems to be scattered on the page.

I also don't know, if the type is valueable in the table. Imagine: You are searching for a setting, then you see a list of settings and the type. Does the type here really help you?

Currently not on my working machine, mostly not before Monday, so can't check.

linawolf:
The table  replaces the contents directive that we have until now, it worked just the same by jumping to the entry:
image
The type might not always but sometimes be helpful in an overview. Also which additional information from the confval should be displayed here is configurable.

brotkrueml:
Another addition: You are mixing again two different scopes in one PR: introducing the confval thingy and moving the examples to the bottom and out of the relevant section. Both should be discussed separately in different PRs as they are not connected.

linawolf:
If you don't feel comfortable reviewing this PR, please let someone else do it. And yes in my eyes they are connected.

PR #1243, opened 2024-05-11
[TASK] cObjects with confval tables

linawolf:


brotkrueml (commented):
See my comment in #1242 about this feature.

PR #1246, opened 2024-05-20
[TASK] Refactor Plugin page

linawolf:
* Use confval list
* link between examples and settings
* reorder that settings and examples are distinct lists, reuse examples where feasable


linawolf:
@froemken How would you feel about restructuring pages like this?

froemken (approved):
Wow...looks much cleaner now. Great work.

PR #1254, opened 2024-05-22
fix typo with includeJS / includeCSSLibs

mavolkmer:
`includeJS` instead of `includeCSSLibs` is demonstrated - links were set correctly


garvinhicking (approved):
Thanks for spotting this and contributing a fix! 👍

Issue #1258, opened 2024-05-23
includeJSLibs/includeJSFooterlibs not matching current behaviour


mavolkmer:
### Current behavior
Currently TYPO3 core does not work as described in documentation of `page.includeJSLibs` / `page.includeJSFooterlibs`
There is some inconsistency in code which is already described in https://forge.typo3.org/issues/103882

|page.include...              	| data.data-xxx / data.foo 	| data-xxx / foo 	|
|--------------	|--------------------------	|----------------	|
| CSS          	| Exception                	| +              	|
| CSSLibs      	| Exception                	| +              	|
| JS           	| +                        	| -              	|
| JSFooter     	| +                        	| -              	|
| JSLibs       	| Exception                	| +              	|
| JSFooterlibs 	| Exception                	| +              	|

Array to String Exception, as a string is expected but an array gets delivered
+: working as expected
-: not working

Thus the documentation needs to be adjusted to show the current state of the code.

### TYPO3 versions

12

### Possible Solution

see PR #1259

garvinhicking:
Closed via #1259 - Thanks a lot for directly addressing this with a PR! :)

PR #1259, opened 2024-05-23
[TASK] match includeJS documentastion to sourcecode

mavolkmer:
As described in #1258 there is some inconsistency between source-code and documentation.
For now documentation should reflect current state of sourcecode until a core-patch is developed and merged

### DIFF
--- a/Documentation/TopLevelObjects/Page/Index.rst
+++ b/Documentation/TopLevelObjects/Page/Index.rst
@@ -473,6 +473,19 @@ Properties
         The optional properties from :ref:`includeJS <setup-page-includejs-array>`
         can be applied.
 
+        Currently one difference between :typoscript:`includeJS` and :typoscript:`includeJSFooterlibs` exists:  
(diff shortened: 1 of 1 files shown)


Documentation/TopLevelObjects/Page/Index.rst
```diff
@@ -486,6 +499,19 @@ Properties
         The optional properties from :ref:`includeJS <setup-page-includejs-array>`
         can be applied.
 
+        Currently one difference between :typoscript:`includeJS` and :typoscript:`includeJSLibs` exists:  
+        There is no :typoscript:`data`-array as optional parameter but all keys not explicitly mentioned as parameters are used as additional attributes - behaviour is the same as in :ref:`includeCSS <setup-page-includecss-array>`..
```
garvinhicking, line 503:
```suggestion
        There is no :typoscript:`data`-array as optional parameter but all keys not explicitly mentioned as parameters are used as additional attributes - behaviour is the same as in :ref:`includeCSS <setup-page-includecss-array>`.
```

garvinhicking:
Thanks so much for the contribution and your intensive work on this!

PR #1261, opened 2024-05-24
[TASK] Add documentation for extbase newRecordStoragePid setting

derhansen:
Add documentation for settings.persistence.classes.[classname].newRecordStoragePid

### DIFF
--- a/Documentation/TopLevelObjects/Plugin.rst
+++ b/Documentation/TopLevelObjects/Plugin.rst
@@ -360,8 +360,11 @@ Demonstrates:
             # Includes 4 sub-levels of the storagePid
             recursive = 4
 
-            T3docs\BlogExample\Domain\Model\Comment {
+            T3docs\BlogExample\Domain\Model\Post {
                 newRecordStoragePid = 43
+            }
+            T3docs\BlogExample\Domain\Model\Comment {
+                newRecordStoragePid = 44
             }
         }
     }
--- a/Documentation/TopLevelObjects/Plugin.rst
+++ b/Documentation/TopLevelObjects/Plugin.rst
@@ -341,7 +341,7 @@ Demonstrates:
 
 ..  _setup-plugin-persistence-recursive-example:
 ..  _setup-plugin-persistence-storagePid-example:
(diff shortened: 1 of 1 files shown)


Documentation/TopLevelObjects/Plugin.rst
```diff
@@ -348,6 +359,10 @@ Demonstrates:
             storagePid = 42
             # Includes 4 sub-levels of the storagePid
             recursive = 4
+
+            Vendor\Extension\Domain\Model\MyModel {
```
mbrodala, line 363: This should be more specifically written for the blog example. Also it is unclear why one would need/set this without mentioning Single Table Inheritance. With only a single model for a single table, this option is not necessary. But if there are multiple models for a table with different record types, then this option is useful to store these on different pages.

mbrodala (changes requested):

Documentation/TopLevelObjects/Plugin.rst
derhansen: I agree with the blog example and have change the PR. 

The setting is however pretty self explaining and I would prefer to not reduce it to single table inheritance only (see see this https://typo3.slack.com/archives/C03AM9R17/p1716279888744799 discussion on slack).

mbrodala: Ah, that's correct indeed. This is not limited to SIT but actually about multiple models in an extension/plugin.

Can you still add another model for clarity?

derhansen: I already did, GitHub does not seem to update it in the commentes section. Change see: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/pull/1261/files#diff-2f734f75a0276d75461dfe447428e2d79a23b45b031da21e9071a3a8b9a02510R363

mbrodala: No, I saw that. I mean a 2nd model.

derhansen: Done

mbrodala (approved):

PR #1265, opened 2024-06-02
[TASK] Fix headline level for prefixLocalAnchors

peterkraume:
This fix is only necessary for the 11.5 branch.
In newer branches `prefixLocalAnchors` isn't present anymore and in older branches, everything is fine.


brotkrueml (approved):

brotkrueml:
@peterkraume Thanks for the fix :+1:

PR #1266, opened 2024-06-04
add example PHP

franzholz:
The userFunc PHP code is not described anywhere, is it?

### DIFF
--- a/Documentation/DataTypes/Index.rst
+++ b/Documentation/DataTypes/Index.rst
@@ -213,6 +213,24 @@ function name
 
         Your\NameSpace\YourClass->reverseString
 
+    ..  code-block:: php
+
+        namespace Your\NameSpace;
+
+        class YourClass{
+            /**
+            * Comment
+            *
+            * @param string $content Empty string (no content to process)
+            * @param array $conf TypoScript configuration
+            * @return string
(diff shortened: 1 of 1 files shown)


froemken:
@linawolf Franz has added 3 further empty lines. So, in total we have 5 empty lines there now. Is that something we want to keep? Is that correct behavior?

linawolf:
I would suggest we put the code in an include file so that it can be linted and cgl can be automatically applied. Would you consider the example itself correct?

froemken:
@linawolf Since TYPO3 11 or 12 the target method gets a third argument `ServerRequestInterface $request`. After that is added the method looks good

froemken (approved):
Approved as discussed with @linawolf 
Current problems with this patch will be resolved with follow-up patch

brotkrueml:
This PR was not mergable IMHO, as it has flaws:

- Namespace should be `MyVendor\MyExtension` as in all other examples in the manual
- CGL is not correct (`{` in class line should be in separate line).
- `@return string` is superfluous
- third parameter `$request` should be added

franzholz:
This cannot be edited any more.

brotkrueml:
This was a comment for my team colleages.

linawolf:
@brotkrueml @froemken and I decided to make a follow up and we will take care of it.

PR #1268, opened 2024-06-05
[TASK] Follow up: add example PHP #1266

linawolf:
Move PHP example to file
Insert third parameter

references #1266
releases: main, 12.4

### DIFF
--- a/Documentation/DataTypes/_FunctionName.php
+++ b/Documentation/DataTypes/_FunctionName.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace MyVendor\MyExtension\SomeNamespace;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+class SomeClass
+{
+    /**
+     * @param string $content Empty string (no content to process)
+     * @param array $conf TypoScript configuration
+     */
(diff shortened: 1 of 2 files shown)


froemken (approved):
thank you for the follow-up. Looks good to me.

Documentation/DataTypes/_FunctionName.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace MyVendor\MyExtension\SomeNamespace;
```
brotkrueml, line 3: Actually, I find the "SomeNamespace" strange as the whole string "MyVendor\MyExtension\SomeNamespace" is the namespace. In other examples we use just "MyVendor\MyExtension" as namespace.

Just my 2 cents after merging, again.

froemken: As over 90% of PHP files stays in sub-directories of `Classes` I prefer the 3-parts-namespace. But yes, I don't like "SomeNamespace". Maybe call it `MyService` or whatever. Having classes directly in `Classes` is very very rare.

PR #1276, opened 2024-06-10
[TASK] Properly document typolink.userFunc usage

### LINKED ISSUE #1275: Outdated information about typolink.userfunc
https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/Functions/Typolink.html#userfunc states:

> The userfunction must return an <A>-tag.

but in reality this will lead to an exception `calling typolink.userFunc resulted in not returning a valid typolink` as the return value must now be an instance of `LinkResultInterface`.

Simple example would be 

```php
return (new LinkResult(LinkService::TYPE_URL, 'https://typo3.org'))
            ->withLinkText('I am a link');
```

docs should also provide an example with TYPE_PAGE

garvinhicking:
According to  https://review.typo3.org/c/Packages/TYPO3.CMS/+/72999 the userFunc now must return an instance of `LinkResultInterface`.

(drive-by fix of a few wrong formattings)

Releases: main, 12.4
Closes #1275

### DIFF
--- a/Documentation/Functions/Typolink.rst
+++ b/Documentation/Functions/Typolink.rst
@@ -561,61 +561,33 @@ userFunc
 
     `$content`
         This contains the object implementing :php:`LinkResultInterface`. Inside your :php:`userFunc()` you
-        can call :php:`$content->getUrl()` to get the URL of a link, or :php:`$content->getLinkText()`
-        to get the title of your link. :php:`$content->getLinkConfiguration()` returns an array
-        with all typolink configuration options, and :php:`$content->getAttributes()`
-        (like :typoscript:`typolink.additionalArguments`) returns current anchor link attributes.
-        :php:`$content->getType()` returns the kind of link that is returned like
-        :php:`LinkService::TYPE_PAGE` (specific pages in your TYPO3 setup) or :php:`LinkService::TYPE_URL`
-        for links to external pages.
+        can call for example:
 
-        See the PHP definition of `LinkResultInterface` for the full list of getters.
+        *  :php:`$content->getUrl()` to get the URL of a link,
+        *  :php:`$content->getLinkText()` to get the title of your link,
+        *  :php:`$content->getLinkConfiguration()` for the array with all typolink configuration options,
+        *  :php:`$content->getAttributes()` returns current anchor link attributes (like :typoscript:`typolink.additionalArguments`),
+        *  :php:`$content->getType()` returns the kind of link that is operated on, like
+           :php:`LinkService::TYPE_PAGE` (specific pages in your TYPO3 setup) or :php:`LinkService::TYPE_URL`
+           for links to external pages.
 
-        Since `LinkResultInterface` is an immutable object, you must use the methods :php:`withLinkText()`
-        and/or :php:`withAttributes()` to return a new object variant, which at the end of your
-        :php:`userFunc` must be returned (see below for an example). In case you do not make any
+        See the PHP definition of :php:`LinkResultInterface` for the full list of getters.
+
+        Since :php:`LinkResultInterface` is an immutable object, you must use the methods :php:`withLinkText()`
+        and/or :php:`withAttributes()` to create a new object variant, which at the end of your
+        :php:`userFunc` must be returned (see below for examples). In case you do not make any
         changes to the object, the function must return the original object.
 
     `$conf`
-        Contains an array of the TypoScript configuration of your :typoScript:`userFunc` parameters.
+        Contains an array of the TypoScript configuration of your :typoscript:`userFunc` parameters.
 
     `$request`
         Contains the PSR-7 request object that allows you to operate on your current frontend
         environment and retrieve things like Site Settings, current Language, current URL,
-        related `ContentObjectRenderer` (`$cObj`) and other aspects,
+        related :php:`ContentObjectRenderer` (:php:`$cObj`) and other aspects,
         see :ref:`TYPO3 request object <t3coreapi:typo3-request>`.
 
-    Give this TypoScript example:
-
-    ..  literalinclude:: _StdWrap/_UserFunc.typoscript
-        :language: typoscript
-        :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
-
-    This would first create a TypoLink :php:`LinkResultInterface` that would resolve to something like
-    `<a href="/en/pages/4711?someKey=someValue" rel="noreferrer">My Link Title</a>`.
-
-    But because `userFunc = MyVendor\SitePackage\UserFunctions\TypoLinkUserFunc->createUserFuncLink`
-    is defined you can now manipulate the generated link to your liking:
-
-    ..  literalinclude:: _StdWrap/_UserFunc.php
-        :language: php
-        :caption: EXT:site_package/Classes/UserFunctions/TypoLinkUserFunc.php
-
-    This class would take the :php:`LinkResultInterface` object, retrieve some data from it,
-    alter it, pass it to a new immutable object and return that. Then this is what finally
-    gets emitted after full processing:
-
-    ..  code-block:: html
-        <a href="/en/pages/4711/event/18"
-           target="_blank"
-           title="Custom: My Link Title"
-           rel="noreferrer">Event Title #18</a>
-
-
-    You can also apply a custom `userFunc` to vital things like the
-    :ref:`lib.parseFunc_RTE.userFunc <t3tsref:parsefunc-userFunc>`
-    routine. This would allow you to modify any kind of link generated from the
-    parsing of the Rich-Text-Editor (RTE).
+    See :ref:`typolink-userfunc-examples` for more details.
 
 ..  index:: typolink; Resource references
 ..  _typolink-resource_references:
--- a/Documentation/Functions/Typolink.rst
+++ b/Documentation/Functions/Typolink.rst
@@ -543,27 +543,66 @@ userFunc
     :name: typolink-userFunc
     :type: :ref:`data-type-function-name`
 
-    This passes the link-data compiled by the typolink function to a user-
-    defined function for final manipulation.
+    As `userFunc` is the last property of `typolink`, all previous properties
+    have already been processed into an object of type `LinkResultInterface`,
+    which is passed as the first parameter to the defined `userFunc` for
+    further manipulation.
 
-    The :php:`$content` variable passed to the user-function (first parameter) is
-    an array with the keys "TYPE", "TAG", "url", "targetParams" and
-    "aTagParams".
+    All `typolink` properties are passed as an array as the second parameter.
 
-    TYPE is an indication of link-kind: mailto, url, file, page
+    The current request object incl. current instance of `ContentObjectRenderer`
+    will be passed as third parameter.
 
-    TAG is the full <A>-tag as generated and ready from the typolink
-    function.
+    As the `userFunc` property must also return an object of type
+    `LinkResultInterface`, it makes sense to use this LinkResult object
+    passed and customise it to suit your needs.
 
-    The actual tag value is constructed like this:
+    ..  code-block:: typoscript
+
+        # Register your UserFunc with TypoScript
+        lib.parseFunc_RTE.userFunc = MyVendor\MySitePackage\UserFunc\ModifyTypoLinkUserFunc->modifyTypoLinkResult
+
+    The following PHP example checks whether the object passed contains a link
+    of type `page` (this is the value behind: `LinkService::TYPE_PAGE`). If yes,
+    the link is opened in a new browser tab, the link title is set to the page
+    title and the link text will be wrapped in :html:`<strong>` tags.
 
     ..  code-block:: php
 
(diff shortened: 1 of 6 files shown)


froemken (approved):

brotkrueml (commented):
I miss the "In addition, typolink.userFunc now receives
**a LinkResult object instead of a <a> string with arguments**,
and needs to return a LinkResult object again." part. Only "TYPE" and "TAG" is passed according to this patch.

garvinhicking:
@brotkrueml 

> I miss the "In addition, typolink.userFunc now receives **a LinkResult object instead of a string with arguments**, and needs to return a LinkResult object again." part. Only "TYPE" and "TAG" is passed according to this patch.

I'm not doing enough integrator work to understand what I should do here, do you have a recommendation what to change?

Would it just mean to replace this:

>   The :php:`$content` variable passed to the user-function (first parameter) is
>    an array with the keys "TYPE", "TAG", "url", "targetParams" and
>    "aTagParams".

with this:

>   The :php:`$content` variable passed to the user-function (first parameter) is
>    an array with the keys "TYPE" and "TAG"

?

brotkrueml:
Don't know, haven't used that. Personally, would try to implement it before documenting. But currently less time.

linawolf (approved):
I would suggest to merge it like that as it is a definite improvement.

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,30 @@ userFunc
     :name: typolink-userFunc
     :type: :ref:`data-type-function-name`
 
-    This passes the link-data compiled by the typolink function to a user-
-    defined function for final manipulation.
+    This passes the link-data compiled by the :typoscript:`typolink` function
+    to a user-defined PHP function for final manipulation.
 
     The :php:`$content` variable passed to the user-function (first parameter) is
```
sfroemkenjw, line 549: The first parameter `$content` to the userfunc is an object of type `LinkResultInterface`, it contains all pre-processed data of the URI you know of `typolink` property like page resolvement, target, class, section and mail handling.

sfroemkenjw: The second parameter `$conf` contains the original TypoScript configuration of the `typolink` property.

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,30 @@ userFunc
...
     an array with the keys "TYPE", "TAG", "url", "targetParams" and
     "aTagParams".
 
-    TYPE is an indication of link-kind: mailto, url, file, page
+    `TYPE` is an indication of link-kind: `mailto`, `url`, `file`, `page`
```
sfroemkenjw, line 553: Yes, there are different kinds of link types, but IMO these should stay part of `typolink.parameter` property.

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,30 @@ userFunc
...
 
-    TAG is the full <A>-tag as generated and ready from the typolink
+    `TAG` is the full `<a>`-tag as generated and processed from the :typoscript:`typolink`
```
sfroemkenjw, line 555: There is no TAG anymore. There is a LinkResult containing all information to build a tag

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,30 @@ userFunc
...
     function.
 
-    The actual tag value is constructed like this:
+    The actual tag value must be returned with an instance of :php:`LinkResultInterface`,
```
sfroemkenjw, line 558: The userfunc will retrieve the pre-processed LinkResult allready. So, the userfunc is just a modificator of the delivered LinkResult which can be returns 1 by 1 or in a modified way.

PS: I don't want to explain how to build up your own LinkResult. Sure, it is possible, but IMO that should NOT be the way to go. It could leed to an unexpected email-link, because the TYPO3 delivered LinkResult was of type INPAGE. Brrr...please, don't do so.

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,30 @@ userFunc
...
+    you can not use plain HTML. Examples:
 
     ..  code-block:: php
 
-        $contents = '<a href="' . $finalTagParts['url'] . '"'
-                   . $finalTagParts['targetParams']
-                   . $finalTagParts['aTagParams'] . '>';
+        // Specific URL string
+        return (new LinkResult(LinkService::TYPE_URL, 'https://typo3.org'))
```
sfroemkenjw, line 564: Instead of creating the LinkResolut on your own I always prefer using the `LinkFactory->buildLinkResult`

sfroemkenjw (changes requested):

froemken:
I will provide a new patch this evening

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,66 @@ userFunc
...
+
+        # Register your UserFunc with TypoScript
+        lib.parseFunc_RTE.userFunc = MyVendor\MySitePackage\UserFunc\ModifyTypoLinkUserFunc->modifyTypoLinkResult
+
+    The following PHP example checks whether the object passed contains a link
+    of type `page` (this is the value behind: `LinkService::TYPE_PAGE`). If yes,
+    the link is opened in a new browser tab, the link title is set to the page
+    title and the link text will be wrapped in :html:`<strong>` tags.
 
     ..  code-block:: php
```
linawolf, line 570: I would suggest to use a literalinclude here: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Code/Codeblocks.html#literalinclude

This way we have the php in its own file, it can be linted and cgls automatically applied. Also please add a caption about where the file is located in a project.

linawolf:
Please run cgl fixes with `make fix`

froemken (commented):
First of all: Thank you for your work.

My original idea was to modify the link title, alt-tag, style and format, but never ever modify the link itself. This patch contains code to build up an internal URL manually. Out of scope of TYPO3 API. For internal linking we should use Site Router API. Sure, that may make the examples more complex.

This patch is better than everything we had before and can be merged. Everything else is solvable with a follow-up, if we want.

garvinhicking:
@froemken Thank you for this encouraging comment.

I agree with you. The scope of modifying the URL is not good at this point, and obfuscates the code here.

I would like to:

- move a whole section to an own page
- leave the first paragraph and then link to the page
- modify base example to only modify simple attributes
- add another example how to build
a simple fresh extern link, and then hint at RouterService and/or linkbuilding

Will do that next week and would be great to get your feedback again ☺️

garvinhicking:
@linawolf I've performed the CGL fixes

@froemken I've now split up the pages like suggested. What do you think?

linawolf (approved):

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,51 @@ userFunc
...
+    return an object implementing the :php:`LinkResultInterface`. The currently calculated
+    typolink is passed as an argument to the :php:`userFunc` as an object of the same type. This allows
+    to return either an enriched link, or a completely new one.
+
+    The detailed execution steps are:
+
+    First, the :typoscript:`typolink` will be created as configured by the specified TypoScript.
+    This will result in an object of Type :php:`LinkResultInterface`. This object receives
+    all of the TypoScript :typoscript:`typolink` configuration as properties, and makes them
+    available via corresponding getters and setters.
```
froemken, line 558: There is at least no "setter" in `LinkResult`. This object is immutable, so it can not be modified. The object contains various `with*` methods to "clone" the existing LinkResult into a new LinkResult with one changed property.

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,51 @@ userFunc
...
+    Then your custom :typoscript:`userFunc` is executed and receives the following arguments
+    (delivered via :php:`$contentObjectRenderer->callUserFunction()`):
+
+    `$content`
+        This contains the object implementing :php:`LinkResultInterface`. Inside your :php:`userFunc()` you
+        can call for example:
+
+        *  :php:`$content->getUrl()` to get the URL of a link,
+        *  :php:`$content->getLinkText()` to get the title of your link,
```
froemken, line 567: This is not the link title. This is the text in between: `<a href="https://obi.de">OBI-wan KenOBI is the link text</a>`

Documentation/Functions/Typolink.rst
```diff
@@ -543,27 +543,51 @@ userFunc
...
+        *  :php:`$content->getAttributes()` returns current anchor link attributes (like :typoscript:`typolink.additionalArguments`),
+        *  :php:`$content->getType()` returns the kind of link that is operated on, like
+           :php:`LinkService::TYPE_PAGE` (specific pages in your TYPO3 setup) or :php:`LinkService::TYPE_URL`
+           for links to external pages.
+
+        See the PHP definition of :php:`LinkResultInterface` for the full list of getters.
+
+        Since :php:`LinkResultInterface` is an immutable object, you must use the methods :php:`withLinkText()`
+        and/or :php:`withAttributes()` to create a new object variant, which at the end of your
+        :php:`userFunc` must be returned (see below for examples). In case you do not make any
```
froemken, line 578: That is perfect! And that's the reason why there is no "setter".

Documentation/Functions/_Examples/Typolink-Example.rst
```diff
@@ -0,0 +1,65 @@
...
+================================
+
+Given this TypoScript example:
+
+..  literalinclude:: ../_StdWrap/_UserFunc.typoscript
+    :language: typoscript
+    :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
+
+This would first create a TypoLink :php:`LinkResultInterface` that would resolve to something like
+:html:`<a href="/en/pages/4711?someKey=someValue" rel="noreferrer">My Link Title</a>`.
```
froemken, line 20:
```suggestion
:html:`<a href="/en/pages/4711?someKey=someValue" rel="noreferrer">My Link Text</a>`.
```

Documentation/Functions/_StdWrap/_UserFunc.php
```diff
@@ -0,0 +1,79 @@
...
+        // This example only operates on TYPE_PAGE for internal
+        // links; the userFunc can of course also react to
+        // types like TYPE_URL (external pages)
+        if ($content->getType() === LinkService::TYPE_PAGE) {
+            $cObj = $this->getContentObjectRenderer($request);
+
+            // Check if typolink.userFunc additional data is set
+            // and act on it
+            if ($conf['eventUid.']['data'] ?? null) {
+                $url = $content->getAttribute('href')
```
froemken, line 31: This may break, because there is no cHash attached to the URI.

Documentation/Functions/_StdWrap/_UserFunc.php
```diff
@@ -0,0 +1,79 @@
...
+                // Example how to utilize the cObj and it's wrapping so
+                // that `<a href=...><strong>(TITLE!)</strong></a>`
+                // would be returned.
+                $linkText = $cObj->stdWrap_dataWrap('<strong>{FIELD:title}</strong');
+            }
+
+            return $content
+                ->withTarget('_blank')
+                ->withAttribute('href', $url)
+                ->withAttribute('title', 'Custom: ' . $cObj->data['title'])
```
froemken, line 49: Nice. That's the way to add the link "title" ;-)

Documentation/Functions/_StdWrap/_UserFuncSimple.php
```diff
@@ -0,0 +1,37 @@
...
+use TYPO3\CMS\Frontend\Typolink\LinkResultInterface;
+
+class TypoLinkUserFunc
+{
+    public function createUserFuncLink(
+        LinkResultInterface $content,
+        array $conf,
+        ServerRequestInterface $request,
+    ): LinkResultInterface {
+
```
froemken, line 18: Remove empty space

froemken (changes requested):
Great work. Thank you. I just found some small issues.

Documentation/Functions/Typolink.rst
garvinhicking: Very true. This was a left over from my rewording, thanks for catching this mistake!

Documentation/Functions/Typolink.rst
garvinhicking: Reworded this.

Documentation/Functions/_Examples/Typolink-Example.rst
garvinhicking: Ack, changed this.

Documentation/Functions/_StdWrap/_UserFunc.php
garvinhicking: IMO for the sake of the example that's ok, I wouldn't want to make this any more complex at this point, would that be acceptable for you?

sfroemkenjw (approved):

froemken (approved):

PR #1280, opened 2024-06-12
Remove marker example from parsefunc.short

froemken:


Documentation/Functions/Parsefunc.rst
```diff
@@ -119,23 +119,22 @@ short
 -----
 
 ..  confval:: short
-    :name: arsefunc-short
```
garvinhicking, line 122: LOL, really? :-D

garvinhicking (commented):
(sadly can't comment on the actual change here, too little integrator knowledge...)

Documentation/Functions/Parsefunc.rst
linawolf: :-D

linawolf:
@froemken is there a changelog entry on when the behaviour changed?

froemken:
@linawolf It has never be changed. I just go back until TYPO3 4.5 and it has never been marker-based:
https://github.com/TYPO3/typo3/blob/TYPO3_4-5/typo3/sysext/cms/tslib/class.tslib_content.php#L4540-L4547

linawolf (approved):

linawolf:
Ok, then we documented this wrong in the first place

PR #1283, opened 2024-06-20
Add Note regarding similar condition.

ischmittis:
Add information, that a similar condition is written in lowercase not camelcase.


linawolf (approved):

linawolf:
Thanks, this is helpful

ischmittis:
@linawolf do you need additional MRs for alle Branches?

PR #1284, opened 2024-06-20
[DOCS] Add condition example on using constants

linawolf:
Releases: main, 12.4

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1324,3 +1324,26 @@ usergroup()
     :ref:`condition-frontend-user` and :ref:`condition-backend-user` instead.
     For migration hints see the
     :ref:`changelog <ext_core:deprecation-100349-1680097287>`.
+
+
+..  _condition-examples:
+
+Examples
+========
+
+..  _condition-examples-constant:
+
+Check if a constant is set to a certain value
+---------------------------------------------
+
+TypoScript constants can be used in conditions with the 
+:ref:`Syntax <t3coreapi:typoscript-syntax-conditions-syntax>` for conditions:
+
+..  code-block:: typoscript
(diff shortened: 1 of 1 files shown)


Documentation/Conditions/Index.rst
```diff
@@ -1324,3 +1324,26 @@ usergroup()
...
+Check if a constant is set to a certain value
+---------------------------------------------
+
+TypoScript constants can be used in conditions with the 
+:ref:`Syntax <t3coreapi:typoscript-syntax-conditions-syntax>` for conditions:
+
+..  code-block:: typoscript
+    :caption: EXT:my_extension/Configuration/TypoScript/setup.typoscript
+
+    [{$tx_my_extension.settings.feature1Enabled} == 1]
```
froemken, line 1345: I love examples. Thank you. What about such information:

Works:
```
# constant
foo = 1

# setup
[{$foo} == 1]
```

Works:
```
# constant
foo = 1

# setup
[{$foo} == '1']
```

Works:
```
# constant
foo = 1

# setup
[{$foo} === 1]
```

Does not work:
```
# constant
foo = 1

# setup
[{$foo} === '1']
```

Works:
```
# constant
foo = bar

# setup
[{$foo} === 'bar']
```

Does not work:
```
# constant
foo = bar

# setup
[{$foo} === bar]
```

Does not work:
```
# constant
foo = bar

# setup
[{$foo} == bar]
```

Works:
```
# constant
foo = bar

# setup
[{$foo} == 'bar']
```

Works:
```
# constant
foo = bar

# setup
[{$foo} == "bar"]
```

froemken (commented):
See my examples in source code

linawolf:
Would you like to make a follow up with more examples? Once we have the example section in we can use more examples

froemken (approved):
I will do follow-up

PR #1294, opened 2024-07-05
[TASK] Clarify constants do not work in constant conditions

linawolf:
releases: main, 12.4
resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/925

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1234,10 +1234,9 @@ site()
     ..  code-block:: typoscript
         :caption: EXT:site_package/Configuration/TypoScript/constants.typoscript
  
-        my.constant = myValue
+        my.constant = my global value
         [traverse(site('configuration'), 'settings/some/setting') == 'someValue']
-          # This works but is rather ugly to rely on
-          other.constant = otherValue
+          my.constant = another value, if condition matches
         [global]
 
 ..  index:: Conditions; siteLanguage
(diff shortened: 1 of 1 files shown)


Documentation/Conditions/Index.rst
```diff
@@ -1348,6 +1358,14 @@ TypoScript constants can be used in conditions with the
         page.10.value = The feature 1 of my_extension is not enabled.
     [END]
 
+..  note::
+    TypoScript constants can be used in frontend TypoScript *setup* conditions,
+    but not in Frontend TypoScript *constants* conditions. At the time of
+    evaluation the constants are not yet available in constants conditions.
+
+    It is, however, possible to use :confval:`site setttings <condition-site>`
```
froemken, line 1366:
```suggestion
    It is, however, possible to use :confval:`site settings <condition-site>`
```

Documentation/Conditions/Index.rst
```diff
@@ -1229,6 +1229,16 @@ site()
...
+    Site settings can also be used in the conditions in TypoScript constants:
+
+    ..  code-block:: typoscript
+        :caption: EXT:site_package/Configuration/TypoScript/constants.typoscript
+ 
+        my.constant = myValue
+        [traverse(site('configuration'), 'settings/some/setting') == 'someValue']
+          # This works but is rather ugly to rely on
+          other.constant = otherValue
+        [global]
```
froemken, line 1241: The problem with conditions which are not defined is: They will be rendered in Fluid as is: `{$my.constant}` instead of just empty value. That's why I prefer using this updated example and in that case nothing is `ugly` anymore:

```suggestion
        my.constant = my global value
        [traverse(site('configuration'), 'settings/some/setting') == 'someValue']
          my.constant = another value, if condition matches
        [global]
```

froemken (approved):

PR #1314, opened 2024-07-30
Mention `//` (double slash) operator

cweiske:
- if: https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/Functions/If.html
- Data getText: https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/Functions/Data.html#data-gettext

### DIFF
--- a/Documentation/Functions/If.rst
+++ b/Documentation/Functions/If.rst
@@ -20,6 +20,10 @@ There is no else property available. The "else" branch of an "if" statement is a
 missing feature. You can implement a workaround by a logic based on the
 :ref:`stdwrap-override-conditions`.
 
+Simple "if empty use different value" conditions for record data can be built
+with the :ref:`TypoScript // (double slash) <data-type-gettext-double-slash>`
+fallback operator.
+
 Also check the explanations and the examples further below!
 
 .. contents::
--- a/Documentation/Functions/Data.rst
+++ b/Documentation/Functions/Data.rst
@@ -35,6 +35,8 @@ code.
 Spaces around the colon (:typoscript:`:`) are irrelevant. The :typoscript:`key` is
 case-insensitive.
 
+..  _data-type-gettext-double-slash:
+
 By separating the value of getText with :typoscript:`//` (double slash) a number of
 codes can be supplied and getText will return the first one, which is not
 empty ("" or zero).


Documentation/Functions/If.rst
```diff
@@ -20,6 +20,9 @@ There is no else property available. The "else" branch of an "if" statement is a
 missing feature. You can implement a workaround by a logic based on the
 :ref:`stdwrap-override-conditions`.
 
+Simple "if empty use different value" conditions can be built with the
+:ref:`TypoScript // (double slash) <data-type-gettext>` fallback operator.
```
brotkrueml, line 24: I suggest to add an anchor to the section and use it here, so people don't have to search for the section. What do you think?

Documentation/Functions/If.rst
```diff
@@ -20,6 +20,9 @@ There is no else property available. The "else" branch of an "if" statement is a
 missing feature. You can implement a workaround by a logic based on the
 :ref:`stdwrap-override-conditions`.
 
+Simple "if empty use different value" conditions can be built with the
```
brotkrueml, line 23:
```suggestion
Simple "if empty use different value" conditions can be build with the
```

brotkrueml (commented):
Thanks for your improvement :+1: I have two remarks.

Documentation/Functions/If.rst
cweiske: "can be built" is correct english.
https://textranch.com/c/can-be-built-or-can-be-build/

froemken:
Hello @cweiske 

yeah...the double slash operator is a pretty awesome feature, but please have a look here:

https://github.com/TYPO3/typo3/blob/v13.2.1/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L3762

The operator is only available for TS property "field". So, currently I can not merge your patch as long as that patch will add it directly into the global gettext-part.

Stefan

cweiske:
I clarified it so that it is clear that this option is for record data only.

linawolf (approved):
Cool Feature! Didnt even know about it!

linawolf (dismissed):
Woops, sorry, no I think @froemken s comment ist still valid here. It is not a property of gettext but of field...

linawolf:
Actually looking at https://github.com/TYPO3/typo3/blob/v13.2.1/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L3791 I would say it is a feature of both field and gettext, what would you say @froemken ?

froemken (approved):

froemken:
@linawolf Right. That feature is there for gettext since ages. Checked that until 9 years ago...and it is still older than 9 years.

PR #1325, opened 2024-09-05
[TASK] Use `0` and `1` for boolean values of config object

brotkrueml:
Setting a TypoScript option to `false` does not really disable the option, as it is casted to bool later on. `true` casts to true, which is correct, but for consistency reasons, we should always use `0` and `1` in TypoScript for boolean values.

See also: https://docs.typo3.org/m/typo3/reference-typoscript/12.4/en-us/DataTypes/Index.html#boolean

Releases: main, 12.4


froemken (approved):
Looks good to me. It's looking a little bit confusing because of `boolean` and 0 and 1, but hey...it's TypoScript ;-)

garvinhicking (approved):

PR #1327, opened 2024-09-08
[TASK] Modernize cObject "USER" examples

brotkrueml:
- Avoid GeneralUtility::makeInstance(), use DI instead (class has to be set public therefore)
- Use newly introduced request attribute to retrieve current page ID as best practise
- Add request where missing as argument and improve docblock
- Streamline namespace to be in line with other examples

Releases: main

### DIFF
--- a/Documentation/ContentObjects/UserAndUserInt/_ExampleListRecords.php
+++ b/Documentation/ContentObjects/UserAndUserInt/_ExampleListRecords.php
@@ -5,15 +5,24 @@ declare(strict_types=1);
 namespace MyVendor\SitePackage\UserFunctions;
 
 use Psr\Http\Message\ServerRequestInterface;
+use Symfony\Component\DependencyInjection\Attribute\Autoconfigure;
 use TYPO3\CMS\Core\Database\ConnectionPool;
-use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer;
 
 /**
  * Example of a method in a PHP class to be called from TypoScript
+ *
+ * The class is defined as public as we use dependency injection in
+ * this example. If you do not need dependency injection, the
+ * "Autoconfigure" attribute should be omitted!
  */
+#[Autoconfigure(public: true)]
 final class ExampleListRecords
 {
+    public function __construct(
+        private readonly ConnectionPool $connectionPool,
+    ) {}
+
     /**
      * Reference to the parent (calling) cObject set from TypoScript
      */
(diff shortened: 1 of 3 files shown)


Documentation/ContentObjects/UserAndUserInt/_ExampleListRecords.php
```diff
@@ -27,24 +36,28 @@ public function setContentObjectRenderer(ContentObjectRenderer $cObj): void
...
      */
     public function listContentRecordsOnPage(string $content, array $conf, ServerRequestInterface $request): string
     {
-        $connection = GeneralUtility::makeInstance(ConnectionPool::class)->getConnectionForTable('tt_content');
+        $connection = $this->connectionPool->getConnectionForTable('tt_content');
         $result = $connection->select(
             ['header'],
             'tt_content',
-            ['pid' => (int)$GLOBALS['TSFE']->id],
+            // The request attribute 'frontend.page.information' has been introduced
```
linawolf, line 50: I would mention such information in the text not as a coment in the code

Documentation/ContentObjects/UserAndUserInt/_Hostname.php
```diff
@@ -2,18 +2,21 @@
...
 final class Hostname
 {
     /**
      * Return standard host name for the local machine
      *
-     * @param  string Empty string (no content to process)
-     * @param  array  TypoScript configuration
+     * @param string Empty string (no content to process)
+     * @param array TypoScript configuration
+     * @param ServerRequestInterface $request The current PSR-7 request object
```
linawolf, line 16: This comment contains no information beyond what the FQN would give you?

brotkrueml: It also does not harm, Maybe not everyone knows what a ServerRequestInterface is.

Documentation/ContentObjects/UserAndUserInt/_ExampleListRecords.php
brotkrueml: This was by intention: Then it is to far away from the context. Ppl tend to just copy/paste without reading text above or below the code snippet. So, they copy the comment with it.

linawolf: If they copy this comment they start wondering what a version switch is as they are then not in the context of the docs anymore

brotkrueml:
Abandoned. Someone feel free to pick it up again.

Documentation/ContentObjects/UserAndUserInt/_Hostname.php
sfroemkenjw: We have 2 request objects in TYPO3. The original TYPO3 Request also known als TYPO3_REQUEST and the Extbase Request (which is dying with each further TYPO3 version). What about:
```suggestion
     * @param ServerRequestInterface $request The TYPO3 request object
```

PR #1328, opened 2024-09-09
[TASK] Modernize cObject "USER" example

brotkrueml:
- Avoid GeneralUtility::makeInstance(), use DI instead (has to be set public therefore)
- Use newly introduced request attribute to retrieve current page ID as best practise

Releases: main


Documentation/ContentObjects/UserAndUserInt/_ExampleListRecords.php
```diff
@@ -5,15 +5,24 @@
...
 use TYPO3\CMS\Core\Database\ConnectionPool;
-use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer;
 
 /**
  * Example of a method in a PHP class to be called from TypoScript
+ *
+ * The class is defined as public as we use dependency injection in
+ * this example. If you do not need dependency injection, the
+ * "Autoconfigure" attribute should be omitted!
```
sfroemkenjw, line 17: What do you think about:

```suggestion
 * This class is called via GeneralUtility::makeInstance. That's why you **should** mark 
 * it as `public: true`. When using dependency injection, like in this example, you
 * **have to** mark the class as `public: true`.
```

brotkrueml: I disagree, there is no need to mark a user function as public as long as you don't use DI.

linawolf:
Do we need backports to 12.4 here?

brotkrueml:
Nope, not compatible with v12.

PR #1330, opened 2024-09-10
Note on not resolving the recursion for “root”

julianhofmann:
The new `\TYPO3\CMS\Core\Domain\Repository\PageRepository::getDescendantPageIdsRecursive()` introduced in v12 exits for `$startPageId=0` via early-return without resolving the recursion.

### DIFF
--- a/Documentation/Functions/Select.rst
+++ b/Documentation/Functions/Select.rst
@@ -146,6 +146,8 @@ recursive
 
     Number of recursive levels for the pidInList.
 
+     **Note:** :typoscript:`recursive` is ignored for *special keyword* :typoscript:`pidInList=root`.
+
 
 ..  _select-orderBy:
 
--- a/Documentation/Functions/Select.rst
+++ b/Documentation/Functions/Select.rst
@@ -146,8 +146,8 @@ recursive
 
     Number of recursive levels for the pidInList.
 
-     **Note:** :typoscript:`recursive` is ignored for *special keyword* :typoscript:`pidInList=root`.
-
+    ..  note::
+        :typoscript:`recursive` is ignored for *special keyword* :typoscript:`pidInList=root`.
 
 ..  _select-orderBy:
 


Documentation/Functions/Select.rst
```diff
@@ -146,6 +146,8 @@ recursive
 
     Number of recursive levels for the pidInList.
 
+     **Note:** :typoscript:`recursive` is ignored for *special keyword* :typoscript:`pidInList=root`.
```
froemken, line 149:
```suggestion
    ..  note::
    :typoscript:`recursive` is ignored for *special keyword* :typoscript:`pidInList=root`.
```

froemken:
Hi @julianhofmann 

thank you for your PR. I had a look into that.

TYPO3 Core removes all pidInList values since 2014:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L4999-L5000

The problem was: That `recursive` was processed before `pidInList`. So we had a lot of performance lost because of recursive queries and afterwards TYPO3 founds a `root` in `pidInList` and throws all collected UIDs away.

You're right, with `getDescendantPageIdsRecursive` the recursive check will be stopped, if an UID=0 appears. You got the 0 because of converting all values of `pidInList` to int, which results to 0 for `root`.

Patch looks good in general, but for `note` we have an adminition, which I would prefer to use: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/Reference/ReStructuredText/Content/Admonitions.html#note

Do you like to update your patch?

Stefan

julianhofmann:
I have adjusted the formatting. My fork was not in sync and still had the old `note` formatting in other places (from which I had copied it).

Optimizing performance is always good. And a `pidInList=root` with `recursive=50`, like we just had, is highly questionable... We stumbled during an upgrade from v11 to v12 - and couldn't figure it out initially. According to TSref it _should_ have worked. I have adjusted the formatting. MeiNF ork was not in sync and still had the old `note` formatting in other places (from which I had copied it).

Optimizing performance is always good. And a `pidInList=root` with `recursive=50`, like we just had, is highly questionable.... We stumbled during an upgrade from v11 to v12 - and just couldn't figure it out at first. According to TSref it _should_ have worked. This sparked our interest in getting to the bottom of it and adding to the documentation

froemken (approved):

PR #1332, opened 2024-09-13
[TASK] Describe htmlTag_stdWrap for hreflang usage

garvinhicking:
See https://review.typo3.org/c/Packages/TYPO3.CMS/+/85049

This documentation should help to allow people to use other custom `<html lang="...">` attribution, which was not clearly enough phrased. We would need a proper documentation entry to properly close the underlying issue.

Releases: main, 12.4

### DIFF
--- a/Documentation/TopLevelObjects/Config.rst
+++ b/Documentation/TopLevelObjects/Config.rst
@@ -557,7 +557,7 @@ Properties of 'config'
         For example it can be used to distinctly set a :html:`lang` attribute,
         that may diverge from the automatically created attribute using the site
         language settings. Normally, :html:`lang` is set to something like `en`
-        or `en`, only indicating the primary language. You may want to use the
+        or `de`, only indicating the primary language. You may want to use the
         :html:`hreflang` site language attribute instead (containing for example
         `en-US`, `en-GB` or `de-AT`), which you can achieve via:
 
--- a/Documentation/TopLevelObjects/Config.rst
+++ b/Documentation/TopLevelObjects/Config.rst
@@ -554,6 +554,19 @@ Properties of 'config'
         Modify the whole :html:`<html>` tag with stdWrap functionality. This can be
         used to extend or override this tag.
 
+        For example it can be used to distinctly set a :html:`lang` attribute,
+        that may diverge from the automatically created attribute using the site
+        language settings. Normally, :html:`lang` is set to something like `en`
+        or `en`, only indicating the primary language. You may want to use the
+        :html:`hreflang` site language attribute instead (containing for example
+        `en-US`, `en-GB` or `de-AT`), which you can achieve via:
+
+        ..  code-block:: typoscript
+            :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
+
+            config.htmlTag_stdWrap.override = <html lang="{siteLanguage:hreflang}">
+            config.htmlTag_stdWrap.override.insertData = 1
+
     ..  confval:: index_descrLgd
         :name: config-index-descrLgd
         :type: :ref:`data-type-integer`
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1233,7 +1233,7 @@ site()
 
     ..  code-block:: typoscript
         :caption: EXT:site_package/Configuration/TypoScript/constants.typoscript
- 
+
         my.constant = my global value
         [traverse(site('configuration'), 'settings/some/setting') == 'someValue']
           my.constant = another value, if condition matches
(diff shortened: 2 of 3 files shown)


brotkrueml (dismissed):

Documentation/TopLevelObjects/Config.rst
```diff
@@ -554,6 +554,19 @@ Properties of 'config'
         Modify the whole :html:`<html>` tag with stdWrap functionality. This can be
         used to extend or override this tag.
 
+        For example it can be used to distinctly set a :html:`lang` attribute,
+        that may diverge from the automatically created attribute using the site
+        language settings. Normally, :html:`lang` is set to something like `en`
+        or `en`, only indicating the primary language. You may want to use the
```
froemken, line 560: like "en" or "en"? Do you mean "de" or "en"?

Documentation/TopLevelObjects/Config.rst
```diff
@@ -554,6 +554,19 @@ Properties of 'config'
...
+        :html:`hreflang` site language attribute instead (containing for example
+        `en-US`, `en-GB` or `de-AT`), which you can achieve via:
+
+        ..  code-block:: typoscript
+            :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
+
+            config.htmlTag_stdWrap.override = <html lang="{siteLanguage:hreflang}">
+            config.htmlTag_stdWrap.override.insertData = 1
```
froemken, line 568: Instead of "override" which is a condition (if content is there, or not) I would prefer to use some more basic stdWrap property:

```suggestion
            config.htmlTag_stdWrap.dataWrap = <html lang="{siteLanguage:hreflang}">
```

Documentation/TopLevelObjects/Config.rst
garvinhicking:
```suggestion
        or `de`, only indicating the primary language. You may want to use the
```

Documentation/TopLevelObjects/Config.rst
garvinhicking: I got the codelines from Georg Ringer. I know too little of Integration with these codes, I am not sure what the reasoning was to use 'override' here, and if it would have benefits. Are you absolutely sure it would work with a dataWrap and has no side effects?

froemken: Both are not working. My example will append an additional html-tag (which is wrong) and the one from Georg will remove all previously added html attributes and will not set the hreflang from site language. Instead it will show the hreflang from Locale object contained in SiteLanguage.

froemken: This is so dump. If you work with:

```
config.htmlTag.attributes.class = specialHtmlTagClass
```

the solution from Georg will overwrite everything. So, this attribute is gone in Georgs solution. I remember some special tag properties from TypoScript. I tried with `encapseLines`, but as in $content `<html lang="en_US">` is no ending `<html>` the `tags` property can not replace anything.

OK. I tries something very different: `externalBlocks` of `parseFunc`. It does not work either as HtmlParser_tags does not have stdWrap properties. So, I can also fixAttrib->set with fixed values.

So, currently only `replacement` seems to be an option:
```
config.htmlTag.attributes.garvin = hallo

config.htmlTag_stdWrap.replacement.10 {
  search = #lang="(.*?)"#i
  replace.dataWrap = lang="{SITE:languages.0.hreflang}"
  useRegExp = 1
}
```

Further problem is, that `{siteLanguage:hreflang}` will automatically be redirected into the sub-object `Locale` of `Language` object. But in Locale object still the original hreflang exists and not the modified version from Site Language object:

https://github.com/TYPO3/typo3/blob/main/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L4044-L4046

So, to get the hreflang property from Site Configuration (without Locale object) is to access all languages and extract the first one. See my example TS from above.

Is there anyone with another idea?

linawolf: `override` makes it already clear, that the tag is indeed overridden. If you want to have it fool proof add a comment that it is overriden and you might have to redefine other parameters. I would not use crazy constructs as most ppl hardly understand stdwrap.

froemken: I have created an issue ticket: https://forge.typo3.org/issues/104987

linawolf:
@garvinhicking @froemken how shall we proceed here?

garvinhicking:
@linawolf I have no clue, I cannot resolve this. We can close/abandon the issue and hope for a shining knight to come along and solve it :D

linawolf:
I would suggest to do so as keeping it perpetually open does not help either and there are already conflicts

PR #1337, opened 2024-09-21
[TASK] Add missing page.meta attribute sub-property

brotkrueml:
The more generic `attribute` sub-property can be used to set the attribute to any arbitrary value. An example is already available for the Open Graph meta tags. Therefore, it is more generic than the `httpEquivalent` sub-property.

Releases: main, 12.4


brotkrueml:
Waiting for #1336 to be merged as it may conflict with this one.

linawolf:
Where do you have the information from? Source?

brotkrueml:
Real world.

Look at the examples at the end of the page, there `attribute` is used.

It took me quite some time to get `og:image:width` working (it only works with `attribute = property` given - without it, an error occurs).

brotkrueml:
@linawolf Can we proceed here?

linawolf:
I am not blocking this, but someone who knows more about it would have to review

brotkrueml:
As there is no interest in this PR, closing it.

garvinhicking:
I nust stumbled across this. Would you like to reopen? I'd review :-)

brotkrueml:
I can't as I deleted the branch. Feel free to add it again.

Documentation/TopLevelObjects/Page/Index.rst
```diff
@@ -591,9 +591,14 @@ Properties
 
         For each key the following sub-properties are available:
 
+        `attribute`
+            Sets the attribute for the meta tag. If it is not defined, the
+            default :html:`name` is used.
```
garvinhicking, line 596: Maybe we could be even more verbose here along the lines of:

The :html:<meta> Tag usually uses attributes like :html:<meta name="..."> (HTML5), but :html:<meta property="..."> is also valid for RDFa metadata like specifyinf OpenGraph properties. By specifying the 'attribute' key, the used attribute can be set to 'name' or 'property', defaulting to 'name'

But current version is already an improvement and I'd be happy with it, too!

garvinhicking (approved):

Documentation/TopLevelObjects/Page/Index.rst
brotkrueml: Please make a follow-up. I won't work on this PR anymore.

garvinhicking: Hm.

garvinhicking:
Probably not going to backport this manually.

linawolf:
Manually backported

PR #1344, opened 2024-09-24
[TASK] Improve stdWrap.formattedDate

brotkrueml:
An example for a custom date/time format derived from a timestamp has been added as a showcase.

Additionally, add a heading for each example to improve readability.

Releases: main, 12.4

### DIFF
--- a/Documentation/Functions/Stdwrap.rst
+++ b/Documentation/Functions/Stdwrap.rst
@@ -910,7 +910,7 @@ formattedDate
     .locale
         A locale other than the locale of the site language.
 
-    ..  rubric:: Examples
+    ..  rubric:: Example: Full German output from a date/time value
 
     ..  code-block:: typoscript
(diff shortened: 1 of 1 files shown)


froemken (approved):
Thank you for your patch.

Some marks:
Normally we have 2 whitespaces indent for TypoScript, but as all other examples in that file have 4 whitespaces, it's OK for me.

I have tested your snippet. For me it's `2023-03-17T04:00:00+0100`. Seems you're living at greenwich timezone ;-)

brotkrueml:
My ddev is set up with UTC.

brotkrueml:
About the indentation: I adopted the indents from the example above.

linawolf:
@froemken I just noticed, that the .editorconfig is incomplete.

Introducing a propper .editorconfig would therefore be a follow up.

linawolf (approved):

PR #1350, opened 2024-10-03
[BUGFIX] Remove unnecessary “data.” in _Default.html

oliverthiele:
Accessing the data does not work with contentElement.data.[property], but directly with contentElement.[property] (Tested in TYPO3 version 13.4.0-dev)


linawolf (approved):

linawolf:
We also need to adjust the site package probably: https://github.com/TYPO3-Documentation/TYPO3CMS-Tutorial-SitePackage-Code/blob/main/Resources/Private/Templates/Pages/Default.html

PR #1368, opened 2024-10-24
[DOC] Fix database-query example

s2b:


froemken:
@s2b Thank you for providing that patch.

PR #1425, opened 2024-11-14
Remove the section for the TLO “temp.” as no unset() is executed anymore

oliverthiele:
Since TYPO3 v12, no unset() is executed when "temp." is parsed. So there is no longer any difference between addressing "temp." by reference or by copy.


fe-hicking (approved):

garvinhicking (approved):

garvinhicking:
Thanks for this!

PR #1428, opened 2024-11-15
fix: avoid backslash in type values

brotkrueml:


brotkrueml:
Fixes:
image
image

Documentation/Functions/Select.rst
```diff
@@ -72,7 +72,7 @@ pidInList
 
 ..  confval:: pidInList
     :name: select_pidInList
-    :type: *list of page\_ids* / :ref:`stdWrap`
+    :type: *list of page uids* / :ref:`stdWrap`
```
garvinhicking, line 75: Maybe "parent page uids" or "page pid's", because the sentence after it talks about pids instead of uids?

brotkrueml: pids are page uids

garvinhicking (approved):

PR #1436, opened 2024-11-20
[BUGFIX] Fix `siteLanguage("locale")` TypoScript condition example

cundd:
The current example for the `siteLanguage` condition does not seem to be working. If I understand the code correctly, the locale string will be normalized to use dashes (e.g. `fr-CH`) as separator instead of an underscore (`fr_CH`) (see [Locale.php](https://git.typo3.org/typo3/typo3/-/blob/22b303f63282e422533b37b09c624900c5995c70/typo3/sysext/core/Classes/Localization/Locale.php#L104).

The output of "locale" in `<f:debug>{siteLanguage}</f:debug>` is also a bit misleading as it uses the `posixFormatted` version instead of the one used in the TypoScript condition.

Please check this proposed change carefully, as I may not understand the whole picture and/or some things depend on the actual hosting environment.

### DIFF
--- a/Documentation/Conditions/Index.rst
+++ b/Documentation/Conditions/Index.rst
@@ -1399,7 +1399,7 @@ siteLanguage()
         Returns the language ID as an integer.
 
     :typoscript:`siteLanguage("locale")`
-        Returns the current locale as a string, for example `en_GB` or `de_DE`.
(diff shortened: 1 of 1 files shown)


linawolf:
Attention! A preport to main is probably needed

linawolf:
If it is not to much work that would be great. Our automatic backport workflow only works for PRs against main

PR #1449, opened 2024-11-27
[BUGFIX] Fix `siteLanguage("locale")` TypoScript condition example

cundd:
The current example for the `siteLanguage` condition does not seem to be working. If I understand the code correctly, the locale string will be normalized to use dashes (e.g. `fr-CH`) as separator instead of an underscore (`fr_CH`) (see [Locale.php](https://git.typo3.org/typo3/typo3/-/blob/22b303f63282e422533b37b09c624900c5995c70/typo3/sysext/core/Classes/Localization/Locale.php#L104).

The output of "locale" in `<f:debug>{siteLanguage}</f:debug>` is also a bit misleading as it uses the `posixFormatted` version instead of the one used in the TypoScript condition.

Please check this proposed change carefully, as I may not understand the whole picture and/or some things depend on the actual hosting environment.

Successor of PR https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/pull/1436


linawolf:
Tests should run after rebasing, failures where not related to this PR

cundd:
I synced the changes from upstream to my fork in Github.

froemken (approved):

froemken:
Hello @cundd 

I have tested that behavior. In TYPO3 core I found a Charset normalizer which converts all underscores to dashes. Thank you for your patch. I have just merged it.

Stefan

PR #1454, opened 2024-11-28
Mention "in" strict comparison for tree.pagelayout

mbrodala:


linawolf:
I wonder: do we have any styleguide etc regarding string delimiters in TypoScript constants? In this example we would now use both single and double quotes

mbrodala:
Not that I'd knew. From TypoScript POV there is no difference at all. Also the Expression Syntax [accepts both](https://symfony.com/doc/current/reference/formats/expression_language.html#supported-literals).

linawolf (approved):

PR #1462, opened 2024-12-04
[TASK] Remove 'type' from note about bad parameters

kitzberger:
In previous versions of TYPO3 this might have been true, but I don't see any reason, why `type` shouldn't be used in `config.linkVars`.

This goes probably back to: https://forge.typo3.org/issues/25317

I've tested it with TYPO3 12 and couldn't find any side effects.

Here's my scenario:

```
pageTest = PAGE
pageTest {
    typeNum = 1234

    config.linkVars = type

    10 = TEXT
    10.typolink.parameter = 1
    10.typolink.returnLast = url
    10.noTrimWrap = |Link: |<br>|
}
 ```

### DIFF
--- a/Documentation/TopLevelObjects/Config.rst
+++ b/Documentation/TopLevelObjects/Config.rst
@@ -632,7 +632,7 @@ Properties of 'config'
 
         ..  note::
 
-            Do **not** include the `type` and `L` parameter in the linkVars
+            Do **not** include the `L` parameter in the linkVars
             list, as this will result in unexpected behavior.
 
     ..  confval:: message_preview


froemken:
Hello @kitzberger 

if you do not work with RouteEnhancer PageType decorator you're just fine, but as most people outside are working with PageType decorator https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Routing/AdvancedRoutingConfiguration.html#routing-pagetype-decorator the $_GET parameter `type` will not be interpreted anymore. The `type` and `L` parameter will be removed in near future. So, it's good to not relate on them.

Thank you for your patch, but I don't think that we will merge that.

Stefan

kitzberger:
@froemken, thanks for explaining.

How would you ensure then to render links within a non-0 typeNum that use their own `type`-parameter? For single typolinks this could be achieved by specifying the type appended like `typolink.parameter = 1,1234`. 

But for generated menus or links placed in RTE, it's only possible by using linkVars=type, isn't it?

froemken:
Hello @kitzberger 

Do you see the `map` part in Page decorator? Only `type` parameters defined here are valid. So, if you set a link to use `type` 123 it will automatically looked up there and implemented with the human readable string into the link. If you write the URL manually somewhere and add `&type=123` and page decorator is used IMO this `type` parameter will not be interpreted by TYPO3 anymore. So, there is no need to set the allowed types in `linkVars`.

kitzberger:
@froemken, oh, you're absolutely right. Thanks for elaborating!

PR #1499, opened 2024-12-27
[TASK] Fuse function numrows into stdwrap

linawolf:
It cannot be used in any other place so it is just a stdwrap function


linawolf:
I think I will solve this another way

PR #1510, opened 2025-01-29
fix: correct display of itemsLimitPerTable

brotkrueml:
The superfluous backtick between `Default` and the description displays the first paragraph of the description in the default section.

Releases: main, 13.4


garvinhicking (approved):

garvinhicking:
Thank you!

PR #1513, opened 2025-01-30
[BUGFIX] Do not pretend to be a cObject

julianhofmann:
`templateName` is of type string/stdWrap.
However, the snippet gives the impression that it is a cObject due to `templateName = TEXT`.

Working:
```
templateName = Default
```

Working
```
templateName.stdWrap.cObject = TEXT
templateName.stdWrap.cObject.value = Default
```

*Not* working
```
templateName= TEXT
templateName.value = Default
```

### DIFF
--- a/Documentation/ContentObjects/Fluidtemplate/_includes/_templateNameAutomatic.typoscript
+++ b/Documentation/ContentObjects/Fluidtemplate/_includes/_templateNameAutomatic.typoscript
@@ -1,6 +1,6 @@
 lib.stdContent = FLUIDTEMPLATE
 lib.stdContent {
-    templateName.stdWrap {
+    templateName {
         cObject = TEXT
         cObject {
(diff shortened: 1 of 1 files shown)


froemken:
Hey @julianhofmann 

good point. In that case we can also remove the `stdWrap` part, right?

```
templateName.cObject = TEXT
templateName.cObject {
```

If you confirm, can you update that snippet within your patch?

Thank you for your help

Stefan

julianhofmann:
Also, that's a good point. ;-) I had the `TEXT`-cObject in mind, where _stdWrap_ on the `stdWrap` property is recommended instead of at the root level.
It's even easier here. That's right.

froemken (approved):

froemken:
Hi @julianhofmann 

Thank you for the update. Merged.

Stefan

PR #1521, opened 2025-02-11
[BUGFIX] Align headline and description in example

pixelbrackets:
Headline says that the following example does not allow deletion, desxription says that this code snippet will allow deletion (which is true).


linawolf (approved):

linawolf:
Thanks for fixing

PR #1525, opened 2025-02-19
[TASK] Display page and config on top level of Frontend TypoScript

linawolf:
image

The page is the starting point for whatever rendering is configured via TypoScript. It is hard to find within the "Top level object."

The TypoScript config also has a central and special handling.

Leave hints of how they can be found in the previous place in the Top Level objects

Releases: main, 13.4, 12.4

### DIFF
--- a/Documentation/TopLevelObjects/Config.rst
+++ b/Documentation/TopLevelObjects/Config.rst
@@ -13,7 +13,7 @@
 `config`: TypoScript configuration
 ==================================
 
-The `config` top-level object does not have to be initialized,  configuration
+The `config` top-level object does not have to be initialized, configuration
 settings can be made like this:
 
 ..  code-block::
--- a/Documentation/TopLevelObjects/Config.rst
+++ b/Documentation/TopLevelObjects/Config.rst
@@ -1,3 +1,5 @@
+:navigation-title: Config
+
 ..  include:: /Includes.rst.txt
 ..  index::
     config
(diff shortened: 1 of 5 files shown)


Documentation/TopLevelObjects/PageAndConfig.rst
```diff
@@ -0,0 +1,52 @@
...
+to subsequent integrators and not compatible with extensions that also make
+settings to the `page` top level object.
+
+TYPO3 does not initialize :typoscript:`page` by default. You must initialize this
+explicitly, for example:
+
+..  code-block:: typoscript
+    :caption: EXT:site_package/Configuration/TypoScript/setup.typoscript
+
+    page = PAGE
```
sandra9967, line 40: The problem I see here is that we have duplicate content. Also in TopLevelObjects/Page/Index.html we have defined the PAGE object. I think it could be useful to focus on one of the spots (TopLevelObjects/Page/Index.html or TopLevelObjects/PageAndConfig.html) where we populate the main information transfer of the PAGE object. For me the intuitive way would be to find more introducing word to the PAGE object in TopLevelObjects/Page/Index.html  and then refer to more advanced information on TopLevelObjects/PageAndConfig.html. So as actual improvement advices: 
1. Paragraph: PAGE object type in TypoScript: Find a tone of voice that refers to people that have no clue what the PAGE object is for and what the main purpose is. For example: "The PAGE object is the basis of every page in the frontend. You need this object to create a page in TYPO3. In TYPO3 it is common practice to use the variable name page as alias for the PAGE object to initialize a new page in TYPO3. When you created a page which you can read more about in chapter .... (link to more specified version of the PAGE object within the top level section) a basic HTML markup will be rendered in the frontend. All the logic is performed by the so called "RequestHandler". It is possible to define other page types but more about this you can read in chapter (link to muliple pages or guidelines with page,json and xml as example page alias types) "
2. Paragraph: PAGE object type in TypoScript: Remove the line of code and only leave it in the TopLevelObjects/PageAndConfig.html. 
3. Transfer the paragraphs "Getting started with the PAGE object", "Multiple pages", "Guidelines", "Properties" ... I think one gets the idea of my intension.

linawolf: Actually the PAGE page only exists once in the source code so there will be no duplicate content. I changed the way it is displayed in the menu but not its location.

We can do more changes to the rst file about the page lateron but to separate concerns I only want to  change its position in the page tree and leave a little information on how to find it with the other top level objects.

Alagts (approved):

Documentation/TopLevelObjects/Config.rst
```diff
@@ -7,38 +9,23 @@
...
-property `config`, actually the array `'config'` is of type CONFIG. It is a
-"top-level-object" because it is not subordinate to any other configuration
-setting.
-
-In PHP you can refer to the array within :file:`typo3/sysext/frontend/Classes/`
-files by writing :php:`$GLOBALS['TSFE']->config['config']`.
-
-This typoscript "top level object" `config` provides access to the internal
-array. This means, that configuration settings can be made easily. For example:
+The `config` top-level object does not have to be initialized,  configuration
```
froemken, line 16:
```suggestion
The `config` top-level object does not have to be initialized, configuration
```

Documentation/TopLevelObjects/Config.rst
```diff
@@ -7,38 +9,23 @@
...
+settings can be made like this:
 
 ..  code-block::
+    :caption: config/sites/my_site/setup.typoscript
 
     # TypoScript
-    config.debug = 1
+    config {
+        headerComment = Made with ❤ by your TYPO3 Documentation Team
```
froemken, line 24: hehe. nice one ;-)

Documentation/TopLevelObjects/PageAndConfig.rst
```diff
@@ -0,0 +1,52 @@
...
+===============================================
+`page` and `config` on the TypoScript top level
+===============================================
+
+By convention the main rendering of a page (page type 0) is configured in a top
+level object called `page`. It is of type
+`PAGE <https://docs.typo3.org/permalink/t3tsref:object-type-page>`_.
+
+The main TypoScript configuration is always done in a top level object called
+`configuration`. It is of type
```
froemken, line 15:
```suggestion
`config`. It is of type
```

PR #1549, opened 2025-03-18
Fix wrong assignment

merzilla:
Assign "PAGE" instead of "10" here.


froemken (approved):

froemken:
Thank you for providing this patch. I have backported it to 13 branch, too.

PR #1551, opened 2025-03-23
typo in TypoScript

SvenJuergens:
With the example above, it should be uids
https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/ContentObjects/Pageview/Index.html#cobj-pageview-data-settings-example


linawolf:
Thanks for catching this!

linawolf (approved):

PR #1566, opened 2025-05-17
setup-page-bodyTag should not point to Properties

franzholz:
If you click on the link setup-page-bodyTag`, then you come to bodyTagCObject back again. This is misleading. You need to come to bodyTag.

bodyTagCObject
        :name: page-bodyTagCObject
        :type: cObject

        This is the default body tag. It is overridden by :ref:`setup-page-bodyTag`


Documentation/TopLevelObjects/Page/Index.rst
```diff
@@ -205,6 +205,7 @@ Properties
         in the future. Therefore you will often see that people use the number
         10 and no number 1 is found.
 
+..  _confval-page-bodytag:
```
linawolf, line 208:
```suggestion
```

linawolf (approved):

PR #1570, opened 2025-05-27
clear the wording

franzholz:
This is not understandable:

"Although you cannot override values TypoScript-style (using the operators and all) the properties of the object which has the reference will be merged with the configuration of the reference."


sarahmccarthy123 (approved):
Looks good :-) I've just made a few minor language corrections. I hope this is ok

PR #1581, opened 2025-06-13
[DOCS] Remove outdated paragraph

FlyKasual:
There is no necessity anymore for user functions to be prepended with `user_` anymore, so that confusing and vague sentence can be removed.


linawolf:
Thanks a lot and congratulations to your first doc edit ;)

PR #1594, opened 2025-07-20
add hint to install extension "setup"

franzholz:
I have search around without success after I missed the TypoScript module on a TYPO3 website and could not find how to activate it.

### DIFF
--- a/Documentation/Guide/FirstSteps/Index.rst
+++ b/Documentation/Guide/FirstSteps/Index.rst
@@ -6,7 +6,7 @@
 First steps
 ===========
 
-The system extension "setup - TYPO3 CMS Setup" must be installed.
+The system extension "tstemplate - TYPO3 CMS TypoScript" must be installed.
 
 The basic rendering is defined in the "Setup" field of the main TypoScript record.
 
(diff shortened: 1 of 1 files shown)


linawolf:
We hear your frustration — getting stuck like this is never a good experience. But to clarify: EXT:tstemplate is enabled in all default TYPO3 installations documented at [docs.typo3.org](https://docs.typo3.org/), including DDEV, Docker, and zip-based setups.

Our documentation is written with the assumption that default system extensions are enabled. Documenting how to enable every core extension that's been manually disabled would create unnecessary complexity for the vast majority of users.

For users working with minimal or custom Composer setups, we recommend using the [Composer Helper](https://get.typo3.org/misc/composer/helper) to ensure required system extensions are installed.

The PR was closed in alignment with these guidelines. We appreciate the intent to help, but we won’t be reopening this one.

franzholz:
This Composer Helper page should be republished also in another format to give an overview over all system extensions. This would then be searchable by users. And also an AI tool could use it. 
Many users like me still do not use composer but the traditional easy way. So we would never look into a composer helper documentation.

PR #1597, opened 2025-08-07
Adjust spelling of second in cache_period section

Julian-Schuierer:


Julian-Schuierer:
Hi Garvin, ah makes sense – thank you for pointing it out! Will do :)

PR #1601, opened 2025-08-25
Do not escape backslashes in TypoScript

cweiske:
.. it does not with with double backslashes in TYPO3v11


sarahmccarthy123 (approved):
Looks good :-)

sarahmccarthy123:
Failing tests are not related to pull request

PR #1610, opened 2025-09-09
Fixed wrong TSconfig syntax

spoonerWeb:


garvinhicking (changes requested):
"main" would help to ease backport

PR #1635, opened 2025-10-17
[BUGFIX] Use correct property `fieldName`

julianhofmann:
`field` would be a property of stdWrap, but not an subproperty of `references`, which is needed in this snippet.

https://docs.typo3.org/permalink/t3tsref:confval-filesprocessor-references-fieldname


linawolf (approved):
Thanks for catching!!

PR #1645, opened 2025-10-25
[FEATURE] Use translation domain mappings over LLL as simplified labels

linawolf:
References: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1370
Releases: main


froemken (changes requested):
Quote: `The LLL: prefix can (should) be omitted when using domains in
LanguageService->sL().`

and 

Quote:
```
If the input contains a colon (:), it will be treated as a
label reference. Examples:

'LLL:EXT:backend/.../locallang_toolbar.xlf:save'
(traditional)
'backend.toolbar:save' (new)
'Save' (plain string, returned as-is)
```

All the changes here are `LLL:backend/`. That's not what the ref describes

PR #1653, opened 2025-11-21
[DOC] Fix incorrect includeCSS usage in TypoScript Page example

HimanshuRamavat07:
The "Loading JavaScript in TypoScript" section incorrectly placed `includeJSFooter` inside `includeCSS`, which is not valid TypoScript. JavaScript includes must be placed under `includeJSFooter`, not `includeCSS`.

This update corrects the example and prevents confusion for integrators.


linawolf (approved):
thanks for catching!

PR #1723, opened 2026-05-12
Substitute some usages of TSFE

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1074
Releases: main, 14.3, 13.4


brotkrueml:
I suggest to merge this also to 13.4 to avoid breakings for users when upgrading to 14.3 (when they adopted the old syntax from the docs).

linawolf (approved):

PR #1725, opened 2026-05-12
Remove example for "Images in lightbox TopUp"

brotkrueml:
The example uses "TSFE : lastImageInfo | origFile" which is not available anymore with TYPO3 v14. The example is also rather complex and uses the TopUp lightbox which was last updated 16 years ago (https://github.com/archan937/topup). Therefore, we should remove the whole example. The same feature can be achieved nowadays more easily with Fluid.

Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1074
Releases: main, 14.3

### DIFF
--- a/Documentation/Functions/Imagelinkwrap.rst
+++ b/Documentation/Functions/Imagelinkwrap.rst
@@ -329,9 +329,8 @@ Example: Use alternative parameters for the a-tag
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 This way it is possible to use a lightbox and to display
-resized images in the frontend. More complete examples are
-:ref:`imageLinkWrap-example-fancybox` and
-:ref:`imageLinkWrap-example-topup`.
+resized images in the frontend. A more complete example is
+:ref:`imageLinkWrap-example-fancybox`.
 
 ..  code-block:: typoscript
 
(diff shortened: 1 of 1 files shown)


Documentation/Functions/Imagelinkwrap.rst
```diff
@@ -527,36 +526,3 @@ and use `fancybox <http://fancybox.net>`_:
           dataWrap = class= "lightbox" data-fancybox-group="lightbox{field:uid}"
        }
     }
-
-..  _imageLinkWrap-example-topup:
```
linawolf, line 531: Please move the permalink to the main section instead of deleting it.

brotkrueml: Hmm, I disagree. The section is not available anymore, so - if someone set a link - this should hard fail instead of linking to a page where this is not covered anymore. IMHO.

linawolf: It is a general rule we do not remove a permalink unless the concept itself was removed from the TYPO3 Core

Documentation/Functions/Imagelinkwrap.rst
```diff
@@ -329,9 +329,8 @@ Example: Use alternative parameters for the a-tag
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
linawolf, line 335:
```suggestion
Needs :typoscript:`JSwindow = 0`.

..  _imageLinkWrap-example-topup:
..  _imageLinkWrap-example:

Example: Use alternative parameters for the a-tag
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

linawolf (approved):

PR #1733, opened 2026-05-14
[TASK] ContentObjectRenderer->$checkPid_badDoktypeList` was deprecated

linawolf:
The property was intended to cache a comma-separated list of page
doctypes that should be excluded from link target checks, but it was
never written to or read from any code path in TYPO3 itself.

References: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1703
Releases: main, 14.3


sarahmccarthy123 (approved):
Looks good :-)

PR #1785, opened 2026-07-30
Uninstall Code Snippet generator

linawolf:
Code snippets will now be maintained in this manual again

Releases: main


linawolf:
This turned out to have reached the release branches as well, through two separate pull requests rather than a backport of this one:

* 13.4 — #1787, a separate pull request with the same content
* 14.3 — #1789, the backport of #1788, which renamed the fragments to `.fluid.html`

Three chapter files lost their heading, their intro and their anchor here and were left as the tail of the chapter: `_HeedTheOrder` (24 lines to 5), `_ModifyTheOrder` (25 to 9) and `_TypoScriptAnArray` (51 to 17). Because the pages switched from `include` to `literalinclude`, the remaining prose renders as code on the published page.

Restored in #1860, which is labelled for both branches. The other 51 files removed here are fine: their four dropped lines are the generator comments and the `code-block` directive, which the including page now supplies.

PR #1795, opened 2026-08-31
Replace wrong explanations from TypoScript reference operator notes and examples

Bunnyfield:
Correct the documentation for the =< reference operator. References are resolved in frontend TypoScript setup wherever the consuming content-object mechanism supports them; tt_content.* is a common and recommended use case, not a technical path restriction.


Bunnyfield:
Just missed the fact that this pointed to the wrong branch, will come up with a new PR for main

PR #245, opened 2019-10-22
Refactor page.meta and carray type

tmotyl:
- moves setup/meta page to setup/page#meta
- removes setup/carray page (the carray type was used inconsistenly
and did more harm than good)
- replace occurrences of ->CARRAY type with property.[array] and cObject
type
- change the header "data types" to "data/value types" as TypoScript syntax
manual refers to the data types using term "value"
- added few examples

PR #1830, opened 2026-09-10
[TASK] Document how multiple condition criteria are combined

linawolf:
Issue 518 asked whether several criteria belong into one pair of square brackets or into several chained ones. The answer was given in the issue but never reached the manual: chaining, the syntax used before TYPO3 v9, was removed and now fails silently, because the tokenizer treats everything after the first closing bracket as a comment. State that, add the missing "and" example next to the existing "or" one, document that "not" binds more strongly than "and" and "and" more strongly than "or", and include the _not.typoscript snippet, which sat unused in the repository while the negation was mentioned without an example.

See: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/issues/518
Releases: main, 14.3, 13.4
Signed-off-by: Lina Wolf
Assisted-by: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_011hMW1LqMyafG9e1Si6eL9W

PR #1900, opened 2026-09-17
[TASK] Show the three examples that were written but never included

linawolf:
Seven snippet files below Documentation were referenced by no page. Four
of them are dead, but the other three were written on purpose in feature
pull requests and only the literalinclude was forgotten, so the text
around them promises an example that never appears.

The browse navigation example returns. The Browse page ended on "The
result of each data processor can then be used", and then stopped. Its
Fluid template gets the .fluid.html extension the manual has used since
#245.

The download presets example is corrected. The page included the page
TSconfig file twice, the second time captioned user.tsconfig, although
the sentence above it says the setting is manipulated "by adding the
page. prefix" - which is exactly what the unused file does and the
included one does not.

The v12 example returns, with a sentence saying what it is for: TYPO3
v12 does not know site sets, so the set folders are registered as static
files behind a version check. Its extension key said "my-extension" with
a hyphen, which is not a key TYPO3 would find; it reads my_extension now,
like every other example in the manual.

The four dead files go: three TypoScript snippets and one template left
behind in CodeSnippets when the snippet generator was uninstalled with
#1785 and the templates were renamed to .fluid.html with #245. Nothing
has pointed at them since.

Releases: main, 14.3, 13.4
Assisted-by: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_012Nmtd7UhUXQQXkSWGjN4ug
Signed-off-by: Lina Wolf


froemken (approved):
Thank you @linawolf! Outstanding cleanup and very attentive catch on the copy-pasted TSconfig download presets and the missing v12 site sets snippet.

Local rendering test with `render-guides` compiled cleanly with 0 warnings.

**Note on merge order regarding `Documentation/Syntax/Conditions/Index.rst`:**
PR #1830 touches this exact line and includes `_codesnippets/_not.typoscript` as well (with the expanded phrasing `"Single criteria can be negated using not or !"` plus precedence docs). Therefore, **PR #1830 should be merged first**, followed by a quick rebase of this PR (which will cleanly drop or resolve that hunk) before merging this one.

